PR #39, opened 2022-02-19
Done  react_people-table-advanced

taras-pyda:
[DEMO LINK](https://taras-pyda.github.io/react_people-table-advanced/)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,164 @@
...
+          </tr>
+        </thead>
+
+        <tbody>
+          {visiblePeople.map(person => {
+            const personUrl = `${person.name.toLowerCase().replace(/ /g, '-')}-${person.born}`;
+
+            return (
+              <tr
+                key={person.name}
```
vitaliy-martynyuk, line 150: not sure whether it is a good idea to set a person.name as a key value, as there might be 2 or even more people with the same names in the list

vitaliy-martynyuk (approved):
well done! 🚀

PR #69, opened 2022-07-11
Develop

DariaVeretyak:
https://dariaveretyak.github.io/react_people-table-advanced/


src/components/NewPerson/NewPerson.tsx
```diff
@@ -0,0 +1,273 @@
...
+          type="text"
+          value={name}
+          required
+          onChange={(event) => {
+            setName(event.target.value);
+            setErrorName(false);
+          }}
+          onBlur={
+            () => getValidationName()
+          }
```
Bohdan-Maksymiuk, line 101:
```suggestion
          onBlur={getValidationName()}
```

src/components/NewPerson/NewPerson.tsx
```diff
@@ -0,0 +1,273 @@
...
+            id="sex-m"
+            name="sex"
+            value="m"
+            checked={sex === 'm'}
+            required
+            onChange={
+              () => {
+                setSex('m');
+                setErrorSex(false);
+              }
```
Bohdan-Maksymiuk, line 123: make custom callback for radio, with will read input value

Bohdan-Maksymiuk (changes requested):
Good job 👍 
* lets add some validation, because I can add such records
Screenshot 2022-07-14 at 11 48 20
* group these radio buttons, they want to be more close to each other
Screenshot 2022-07-14 at 11 48 06

PR #101, opened 2022-09-12
Task solution

sergii-nosachenko:
[DEMO LINK](https://sergii-nosachenko.github.io/react_people-table-advanced/)


src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <input
+            data-cy="NameFilter"
+            type="search"
+            className="input"
+            placeholder="Search"
+            value={query}
+            onChange={(event) => setSearchParams(
+              getSearchWith(searchParams, {
+                query: event.target.value || null,
+              }),
```
dellachaise, line 57: Better create separate function and use it here.

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,118 @@
...
+      </div>
+
+      <div className="panel-block">
+        <div className="level is-flex-grow-1 is-mobile" data-cy="CenturyFilter">
+          <div className="level-left">
+            {['16', '17', '18', '19', '20'].map(century => (
+              <SearchLink
+                params={{
+                  centuries: centuries.includes(century)
+                    ? centuries.filter(c => c !== century)
```
dellachaise, line 74: Try to create another variable name, not `c`. Use one letter as variable name - is a bad practice.

src/components/SortLink/SortLink.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const SortLink: React.FC<TProps> = ({ field }) => {
+  const [searchParams] = useSearchParams();
+  const sort = searchParams.get('sort') || '';
+  const order = searchParams.get('order') || '';
+
+  return (
+    <SearchLink
+      params={{
+        sort: sort !== field || (sort === field && !order)
```
dellachaise, line 18: You could create a variable for this condition and use it here. This way would be easier to reed code.

src/pages/People.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const filteredPeople = filterPeople(people, searchParams);
+  const sortedPeople = sortPeople(filteredPeople, searchParams);
+
+  return (
+    <>
+      <h1 className="title">People Page</h1>
+
+      <div className="block">
+        <div className="columns is-desktop is-flex-direction-row-reverse">
+          {people.length > 0 && (
```
dellachaise, line 55: This condition would be enough. The same in the code lower. Why don't you use it?
```suggestion
          {people.length && (
```

dellachaise (changes requested):
Great job 👍 

Just add a few little changes 👇

src/pages/People.tsx
sergii-nosachenko: @dellachaise Not, it not the same. If we use `!people.length` we will literally get **boolean type**. But in case of `people.length` we will get **number type**. That's why when our array will be empty we will have number 0 rendered on page. You can try and check it)

src/components/PeopleFilters/PeopleFilters.tsx
sergii-nosachenko: @dellachaise I have simple array here used only in one place and I compare element from centuries array with variable century. This is more than self-describle. But if you insist on this - I can change it. But I think it will have no practical meaning here.

sergii-nosachenko:
@dellachaise Fixed some of your comments, but others are controversial and I want to discuss.

src/components/PeopleFilters/PeopleFilters.tsx
dellachaise: You could leave this way. In this case, usage could be different from project to project and in the future, you would do it according to the tools of your project.

sergii-nosachenko: @dellachaise I understand it, of course.

src/pages/People.tsx
dellachaise: Ok, I understand you 👍  I think you could leave it the way you have or try `!!`.

dellachaise (approved):
Well done 👍

PR #154, opened 2023-01-15
people table advanced

Mikitkaaist:
[DEMO LINK](https://Mikitkaaist.github.io/react_people-table-advanced/)


src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,27 @@
+import classNames from 'classnames';
+import { FC } from 'react';
+import { Link, useLocation } from 'react-router-dom';
+import { Person } from '../../types';
+
+type Props = {
+  person: Partial<Person>;
```
vlad-demchuk, line 7: use `?` instead of `Partial`

vlad-demchuk (approved):

PR #159, opened 2023-01-19
people table was implemented

Oleksandr-Filo:
[DEMO LINK](https://Oleksandr-Filo.github.io/react_people-table-advanced/)


shyptia (approved):
Looks good to me🔥

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,125 @@
...
+  }, [query, searchParams]);
+
+  const visiblePeople = useMemo(() => getVisiblePeople(
+    people,
+    sex,
+    query,
+    centuries,
+    sort,
+    order,
+  ), [people, sex, query, centuries, sort, order]);
```
solaryasha, line 67: The function that takes 3 or more arguments is hard to maintain. It's really easy to mess up the order of params, so you should pass an object.

solaryasha: Also, it's better to create a custom hook that returns the result of useMemo, than create a function and call it inside useMemo hooks.

```
export const useVisiblePeople = (params) => {
  return useMemo(() => {
   // filter logic
}, [bla])
}

File: src/pages/PeoplePage.tsx [hunk 2]
```diff
@@ -0,0 +1,125 @@
...
+  useEffect(() => {
+    loadPeople();
+  }, []);
+
+  const sex = searchParams.get('sex');
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries');
+  const sort = searchParams.get('sort');
+  const order = searchParams.get('order');
+
```
solaryasha, line 51: Also, you could use custom hook which takes searchParams object and returns all these fields in the object

solaryasha (changes requested):

src/pages/PeoplePage.tsx
Oleksandr-Filo: done

src/pages/PeoplePage.tsx
Oleksandr-Filo: done

src/App.tsx
```diff
@@ -1,20 +1,19 @@
-import { PeoplePage } from './components/PeoplePage';
+import React from 'react';
+import { Outlet } from 'react-router-dom';
 import { Navbar } from './components/Navbar';
 
 import './App.scss';
 
-export const App = () => {
+export const App: React.FC = React.memo(() => {
```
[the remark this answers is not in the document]
Oleksandr-Filo: fixed

solaryasha (approved):

PR #168, opened 2023-02-02
Added main roating

romanblyashuk94:
[DEMO LINK](https://romanblyashuk94.github.io/react_people-table-advanced/)


GabrielMaksimovich (approved):
GJ!Nothing to add!

src/utils/getSortedPeople.ts
```diff
@@ -0,0 +1,34 @@
...
+      sortedPeoople.sort((a, b) => a.sex.localeCompare(b.sex));
+      break;
+
+    case 'born':
+      sortedPeoople.sort((a, b) => a.born - b.born);
+      break;
+
+    case 'died':
+      sortedPeoople.sort((a, b) => a.died - b.died);
+      break;
```
GabrielMaksimovich, line 25: Maybe we can combine this four cases in two cases, >> sortedPeople.sort(a,b => a[sort].localeCompare(b[sort]), but it's just comment.

romanblyashuk94: agree.
good point. tanks

src/components/Loader/MainNavLink/MainNavLink.tsx
```diff
@@ -0,0 +1,20 @@
+import cn from 'classnames';
+import { FC } from 'react';
+import { NavLink } from 'react-router-dom';
+
+interface Props {
+  to: string,
```
solaryasha, line 6: It's better to use `To` interface from `react-router-dom`, cause it could contain Path interface

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,53 @@
...
+  const onSearchInputChange = (
+    event: React.ChangeEvent<HTMLInputElement>,
+  ) => {
+    setSearchParams(getSearchWith(
+      searchParams,
+      { query: event.target.value || null },
+    ));
+  };
+
+  const CENTURIES = ['16', '17', '18', '19', '20'];
```
solaryasha, line 20: You should:
wrap it with `useMemo` or move it from the component

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }, []);
+
+  const query = searchParams.get('query')?.toLowerCase() ?? '';
+  const selectedSex = searchParams.get('sex');
+  const selectedCenturies = useMemo(() => {
+    return searchParams.getAll('centuries');
+  }, [searchParams]);
+
+  const visiblePeople = useMemo(() => {
+    return getFilteredPeople(people, query, selectedCenturies, selectedSex);
```
solaryasha, line 33: The function that takes 3 or more arguments is hard to maintain. It's really easy to mess up the order of params, so you should pass an object.

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }, [people, query, selectedCenturies, selectedSex]);
```
solaryasha, line 34: Also, it's better to create a custom hook that returns the result of useMemo, than create a function and call it inside useMemo hooks.

```
export const useVisiblePeople = (params) => {
  return useMemo(() => {
   // filter logic
}, [bla])
}

File: src/utils/getSortedPeople.ts [hunk 6]
```diff
@@ -0,0 +1,28 @@
+import { Person } from '../types';
+
+export const getSortedPeople = (
```
solaryasha, line 3: Where do we use this function ?

src/utils/getPreperedPeople.ts
```diff
@@ -0,0 +1,19 @@
+import { Person } from '../types';
+
+const getPersonByName = (
+  people: Person[],
+  searchingPersonName: string,
+) => people.find(person => person.name === searchingPersonName);
+
+export const getPreperedPeople
```
solaryasha, line 8: Where do we use this function ?

solaryasha (changes requested):

src/utils/getPreperedPeople.ts
romanblyashuk94: image
here

src/utils/getPreperedPeople.ts
```diff
@@ -0,0 +1,19 @@
...
+  = (people: Person[]): Person[] => people.map(person => {
+    return {
+      ...person,
+      mother: person.motherName
+        ? getPersonByName(people, person.motherName)
+        : undefined,
```
valerii-derkach, line 14: better to use null

valerii-derkach (approved):
Great!

PR #170, opened 2023-02-03
Develop

lefteris-L:
[DEMO LINK](https://lefteris-l.github.io/react_people-table-advanced/)


src/components/PeoplePage.tsx
```diff
@@ -1,35 +1,116 @@
...
+  };
 
-              <p>There are no people matching the current search criteria</p>
+  useEffect(() => {
+    fetchPeople();
+  }, []);
 
-              <PeopleTable />
-            </div>
+  switch (true) {
```
meforoma, line 65: this is a bit weird `switch(true)`, just not a pattern you could often see
i get the idea,
yet i'd rather nest the conditional rendering logic inside return statement
```
const renderMainElement = !isLoading
  && !hasError
  && people.length

return (
  {isLoading && <Loader />}
  {hasError && ...Something went wrong}
  {!people.length && ...There are no people on the server}
  {renderMainElement && ...}
```

src/components/PeoplePage.tsx
```diff
@@ -1,35 +1,116 @@
...
+
+    default: visibleElement = (
+      <div className="columns is-desktop is-flex-direction-row-reverse">
+        <div className="column is-7-tablet is-narrow-desktop">
+          <PeopleFilters />
+        </div>
+
+        <div className="column">
+          <div className="box table-container">
+            {visiblePeople.length === 0
```
meforoma, line 92:
```suggestion
            {!visiblePeople.length
```

meforoma (approved):
good one ✅ 
consider a few comments of mine

lefteris-L:
> good one ✅ consider a few comments of mine

Thanks, will do

PR #175, opened 2023-02-05
Develop

oleksii-bieliaiev:
[DEMO LINK](https://oleksii-bieliaiev.github.io/react_people-table-advanced/)


src/components/PeopleTable.tsx
```diff
@@ -1,683 +1,233 @@
...
-          <td>Petrus de Milliano</td>
-        </tr>
+export const PeopleTable: FC<Props> = ({ listOfPeople }) => {
+  const { selectedSlug = '' } = useParams();
+  const [searchParams] = useSearchParams();
+
+  const sort = searchParams.get('sort');
+  const order = searchParams.get('order');
+  const query = searchParams.get('query') || '';
+  const sex = searchParams.get('sex');
```
Anastasiia-Svintsova, line 19: Would be good to create an enum for the params.

src/components/PeopleTable.tsx
```diff
@@ -1,683 +1,233 @@
...
+      if (order === 'desc') {
+        [personA, personB] = [personB, personA];
+      }
+
+      switch (sort) {
+        case 'name':
+        case 'sex':
+          return personA[sort].localeCompare(personB[sort]);
+        case 'born':
+        case 'died':
```
Anastasiia-Svintsova, line 49: Same about en enum for these values

src/components/PeopleTable.tsx
```diff
@@ -1,683 +1,233 @@
...
+        person.name.toLowerCase().includes(query.toLowerCase())
+        || person.motherName?.toLowerCase().includes(query.toLowerCase())
+        || person.fatherName?.toLowerCase().includes(query.toLowerCase())
+      ));
+    }
+  };
+
+  filterListByPeopleFilters();
+
+  sortOptions(copyList);
```
Anastasiia-Svintsova, line 79: Could you use hooks for it? It can lead to a memory leak. Call these functions only when you need to (useEffect). And wrap them into useCallback() if you will use them in the dependency array

src/components/PeopleTable.tsx
```diff
@@ -1,683 +1,233 @@
...
+
+    if (sex) {
+      copyList = copyList.filter(person => (person.sex === sex));
+    }
+
+    if (query) {
+      copyList = copyList.filter(person => (
+        person.name.toLowerCase().includes(query.toLowerCase())
+        || person.motherName?.toLowerCase().includes(query.toLowerCase())
+        || person.fatherName?.toLowerCase().includes(query.toLowerCase())
```
Anastasiia-Svintsova, line 72: Don't repeat yourself. Create a method for this action `toLowerCase().includes(query.toLowerCase())`. You do it 3 times.

src/components/PeopleTable.tsx
```diff
@@ -1,683 +1,233 @@
...
+      ));
+    }
+  };
+
+  filterListByPeopleFilters();
+
+  sortOptions(copyList);
+
+  const linkParams = (value: string) => {
+    if (value === sort && order === 'desc') {
```
Anastasiia-Svintsova, line 82: Also create an enum for the order types

Anastasiia-Svintsova (changes requested):
Good job! Fix some comments

agniya-i (approved):

PR #176, opened 2023-02-06
solution v0.1

nzrii-kysel:
[DEMO LINK](https://nzrii-kysel.github.io/react_people-table-advanced/)


src/components/SortLink/SortLink.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <Link
+      to={{
+        search: getSearchWithSort(sort),
+      }}
+    >
+      <span className="icon">
+        <i
+          className={classNames(
```
anna-inozemets, line 45: Pay attention for this classNames. For the first click we should see fa-sort-up, not fa-sort-down, you can see at the demo link

src/utils/getFilteredPeople.ts
```diff
@@ -0,0 +1,48 @@
...
+          return person.sex === 'm';
+        case 'f':
+          return person.sex === 'f';
+        default:
+          return true;
+      }
+    });
+  }
+
+  if (centuries.length) {
```
anna-inozemets, line 35: I guess, in this case we should take just born cuntury into account, you can see the difference between that filter and filter on demo link

anna-inozemets (approved):
Good and clear code!)

src/components/SortLink/SortLink.tsx
nzrii-kysel: i was intentional, I believe the way I did it makes more sense

src/utils/getFilteredPeople.ts
nzrii-kysel: you are right, thanks

src/App.tsx
```diff
@@ -1,20 +1,27 @@
...
-      <div className="section">
-        <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
-        </div>
+    <div className="section">
+      <div className="container">
+        <Routes>
+          <Route path="/" element={<h1 className="title">Home Page</h1>} />
```
Anna-Tkachenko, line 14: let's create separate component for the element

src/App.tsx
```diff
@@ -1,20 +1,27 @@
...
+          <Route path="/home" element={<Navigate to="/" replace />} />
+
+          <Route path="/people">
+            <Route index element={<PeoplePage />} />
+            <Route path=":person" element={<PeoplePage />} />
+          </Route>
+
+          <Route path="*" element={<h1 className="title">Page not found</h1>} />
```
Anna-Tkachenko, line 22: same here

src/components/Navbar/NavbarLink/NavbarLink.tsx
```diff
@@ -0,0 +1,35 @@
...
+  children?: React.ReactNode,
+};
+
+export const NavbarLink: React.FC<Props> = React.memo(({
+  pathname,
+  preserveParams = false,
+  children,
+}) => {
+  const location = useLocation();
+  const search = preserveParams ? location.search : '';
```
Anna-Tkachenko, line 17: 🔥

Anna-Tkachenko (approved):
Well Done! ⭐️
Clean and great solution!

PR #182, opened 2023-02-19
Develop

vitaliidox:
[DEMO LINK](https://vitaliidox.github.io/react_people-table-advanced/)


src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,193 @@
...
+      return { centuries: null };
+    }
+
+    if (!arrParams.includes(sortName)) {
+      arrParams.push(sortName);
+
+      return { centuries: arrParams };
+    }
+
+    arrParams.splice(arrParams.indexOf(sortName), 1);
```
TarasHoliuk, line 23: why do you need it?

src/utils/flteredPeopleList.ts
```diff
@@ -0,0 +1,46 @@
+import { Person } from '../types';
+
+export const filteredPeople: React.FC<Person> = (currentPeople, search) => {
```
TarasHoliuk, line 3:
```suggestion
export const getFilteredPeople: React.FC<Person> = (currentPeople, search) => {
```
`filteredPeople` sounds like it is already the array of filtered people
use verbs in function names

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const addParentsData = (arrayOfPeople: Person[] | undefined) => {
+    if (!arrayOfPeople) {
+      return null;
+    }
+
+    return arrayOfPeople.map((item) => {
+      const person = { ...item };
+
+      if (item.fatherName || item.motherName) {
+        for (let i = 0; i < listOfPeople().length; i += 1) {
```
TarasHoliuk, line 62: you call `listOfPeople()` a few times
no need to create a copy every time. You may do it once
but in this case, you actually do not change the array at all

also, you can use `forEach` to iterate through array

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,127 @@
...
+          }
+        }
+      }
+
+      return person;
+    });
+  };
+
+  const currentPeopleList = addParentsData(getSortedPeopleList()
+    ?.filter((e) => filteredPeople(e, searchParams)));
```
TarasHoliuk, line 78: what is `e` - a bad name

TarasHoliuk (changes requested):

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,127 @@
...
+      .then((data) => setPeopleList(data))
+      .catch((error) => setIsError(error.message))
+      .finally(() => setIsLoad(false));
+  };
+
+  const sortPerson = (a: Person, b: Person) => {
+    switch (sortParam) {
+      case 'name': return a.name.localeCompare(b.name);
+      case 'sex': return a.sex.localeCompare(b.sex);
+      case 'born': return a.born - b.born;
```
OleziO, line 34: Same case.

src/components/PeoplePage/PeoplePage.tsx
OleziO, line 62: You don't need to call `listOfPeople()` multiple times. Just create variables
`cont list = listOfPeople();` and work with `list`

src/components/PeoplePage/PeoplePage.tsx
OleziO, line 78: Rename `e` to `person`

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,127 @@
...
+                <Loader />
+              )}
+
+              {isError && (
+                <p data-cy="peopleLoadingError" className="has-text-danger">
+                  {isError}
+                </p>
+              )}
+
+              {peopleList?.length === 0 && (
```
OleziO, line 108: Replace with `!peopleList?.length`, it will be the same as `peopleList?.length === 0`. Check other similar cases.

src/utils/flteredPeopleList.ts
OleziO: Not resolved

OleziO (changes requested):
Fix all comments

src/components/PeopleFilters/PeopleFilters.tsx
vitaliidox: When all the buttons of the centuries are checked, this part of the code helps to get the correct array of values from the active buttons after deselecting any of them

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const addParentsData = (arrayOfPeople: Person[] | undefined) => {
+    if (!arrayOfPeople) {
+      return null;
+    }
+
+    return arrayOfPeople.map((item) => {
+      const person = { ...item };
+
+      if (item.fatherName || item.motherName) {
+        list.forEach((itemList) => {
```
TarasHoliuk, line 65:
```suggestion
        list.forEach((listItem) => {
```

TarasHoliuk (approved):

PR #183, opened 2023-02-19
commit message

11Olya112:


Anastasiia-Svintsova (changes requested):
- You should fix this test. While people are loading I can't see the loader now. But I see this message instead 
image

https://user-images.githubusercontent.com/99403198/220088858-a7235bb5-5640-4fff-954a-180e14a692f5.mov

- Filters don't work. Fix it, please 

https://user-images.githubusercontent.com/99403198/220089116-d78f93f6-9c1f-4a09-9c58-d0cee73abdbb.mov

- Some of the mothers' and fathers' names also should be links. Check the working version 
image

- There shouldn't be this message if there are people at the table 
image

- It seems that there is a memory leak. There are a lot of rerendeers

https://user-images.githubusercontent.com/99403198/220090403-1135673d-1943-4eb8-a65e-ce306e2e0f9c.mov

- The sorting is not implementing 
image

PR #184, opened 2023-02-21
done

panianchuk:
https://panianchuk.github.io/react_people-table-advanced/#/


src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,133 @@
...
+
+  const [searchParams, setSearchParams] = useSearchParams();
+  const query = searchParams.get('query') || '';
+  const sex = searchParams.get('sex') || '';
+  const sort = searchParams.get('sort') || '';
+  const order = searchParams.get('order') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+
+  useEffect(() => {
+    async function fetchPerson() {
```
mykhalenych, line 26: move from useEffect, and call only fetchPerson()

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,133 @@
...
+      } finally {
+        setIsLoading(false);
+        setIsPeopleLoaded(true);
+      }
+    }
+
+    fetchPerson();
+  }, []);
+
+  let visiblePeople;
```
mykhalenych, line 46:
```suggestion
const visiblePeople = order ? getPeopleShow().reverse() : getPeopleShow();
```
create after getPeopleShow

mykhalenych (changes requested):

DarkMistyRoom (changes requested):
the search input doesn't work while the sex param is chosen
image

And try to pass all tests

DarkMistyRoom (approved):
you render people in reverse order from test expectations.
expectation:
image
your variant:
image

I approve the task, but you may fix it to pass the tests

PR #185, opened 2023-02-22
Solved

Olek-Lytvynenko:
[Demo](https://olek-lytvynenko.github.io/react_people-table-advanced/)


SanyaBratashchuk (changes requested):
image
image
Search doesn't work as expected

src/components/PeopleFilters.tsx
```diff
@@ -27,66 +63,51 @@ export const PeopleFilters = () => {
...
-            </a>
+            {centuriesParams.map(century => (
+              <SearchLink
+                data-cy="century"
+                key={century}
+                params={{
+                  centuries: centuries.includes(century)
+                    ? centuries.filter(number => number !== century)
+                    : [...centuries, century],
+                }}
```
loralevitska, line 74: Move this expressions to variable

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,116 @@
...
+    if (sort === value && !order) {
+      return { sort: value, order: 'desc' };
+    }
+
+    return { sort: null, order: null };
+  };
+
+  const sortPeople = persons.sort((first, second) => {
+    switch (sort) {
+      case 'name':
```
loralevitska, line 38: [Try not to hardcode these values, but use ](https://www.typescriptlang.org/docs/handbook/enums.html)[Enum](https://www.typescriptlang.org/docs/handbook/enums.html) instead

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { Loader } from '../components/Loader';
+import { PeopleTable } from '../components/PeopleTable';
+import { Person } from '../types';
+import { getPeople } from '../api';
+
+export const PeoplePage = () => {
+  const [peoples, setPeoples] = useState<Person[]>([]);
+  const [isLoading, setIsloading] = useState(false);
+  const [isErorrOnLoadPeople, setIsErorrOnLoadPeople] = useState(false);
+  const [isNoPeopleOnServer, setIsNoPeopleOnServer] = useState(false);
```
loralevitska, line 13: Don't add extra variables to state. Use people.length === 0 instead isNoPeopleOnServer

loralevitska (changes requested):
Good job! There are things that should be fixed:

1. Search still not fixed 
image

anstsot (changes requested):
Good job but let's fix last items:

1. it's better to hide table if there are no data (please check the working example)

image

2. according to the task you have to filter people by name, motherNmae or fatherName
image
so you can just roll back previous working code.

Everything else looks fine from my side

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,117 @@
...
+      case SortType.Sex:
+        return order === 'desc'
+          ? second[sort].localeCompare(first[sort])
+          : first[sort].localeCompare(second[sort]);
+      case SortType.Born:
+      case SortType.Dies:
+        return order === 'desc'
+          ? second[sort] - first[sort]
+          : first[sort] - second[sort];
+      default: return 0;
```
loralevitska, line 49:
```suggestion
    switch (sort) {
      case SortType.Name:
      case SortType.Sex:
        return order === 'desc'
          ? second[sort].localeCompare(first[sort])
          : first[sort].localeCompare(second[sort]);
          
      case SortType.Born:
      case SortType.Dies:
        return order === 'desc'
          ? second[sort] - first[sort]
          : first[sort] - second[sort];
          
      default: return 0;
```

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,117 @@
...
-              href="#/people/maria-sturm-1835"
-            >
-              Maria Sturm
-            </a>
-          </td>
-          <td>
-            <a href="#/people/carolus-haverbeke-1832">Carolus Haverbeke</a>
-          </td>
-        </tr>
+        {sortPeople.map(person => (
```
loralevitska, line 89: Better destructure person

loralevitska (approved):
👍

PR #186, opened 2023-02-23
solution

igor-kashchenko:
[DEMO LINK](https://igor-kashchenko.github.io/react_people-table-advanced/)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,180 @@
...
+      if (sort === sortBy && !isReversed) {
+        return 'fas fa-sort-up';
+      }
+
+      if (sort === sortBy && isReversed) {
+        return 'fas fa-sort-down';
+      }
+
+      return 'fas fa-sort';
+    };
```
Avichez, line 68: As a suggestion, we can use here also `classnames`.
```suggestion
    const handleSortClass = classnames('fas', {
        'fa-sort': sort !== sortBy,
        'fa-sort-up': sort === sortBy && !isReversed,
        'fa-sort-down': sort === sortBy && isReversed,
    });
```

Avichez (approved):
Great solution 🔥

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,52 @@
...
+        </SearchLink>
+
+        <SearchLink
+          className={classNames({
+            'is-active': sex === 'f',
+          })}
+          params={{ sex: 'f' }}
+        >
+          Female
+        </SearchLink>
```
Anna-Tkachenko, line 49: will be great to create universal component for these filters

src/components/PeopleFilters.tsx
```diff
@@ -27,66 +69,57 @@ export const PeopleFilters = () => {
...
-            </a>
-
-            <a
-              data-cy="century"
-              className="button mr-1 is-info"
-              href="#/people?centuries=19"
-            >
-              19
-            </a>
+            {['16', '17', '18', '19', '20'].map(century => {
```
Anna-Tkachenko, line 72: let's move it to separate variable

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+    const sortedPeople = useMemo(() => (
+      getSortedPeople(visiblePeople, sort, isReversed)
+    ), [visiblePeople, sort, isReversed]);
+
+    if (!sortedPeople.length) {
+      return <p>There are no people matching the current search criteria</p>;
+    }
+
+    const updateSearchParams = (sortBy: SortBy) => {
```
Anna-Tkachenko, line 37: will be better to move it to some helper function instead of create this function inside the component

src/components/PeopleTable/PeopleTable.tsx
Anna-Tkachenko: agree, check @Avichez comment

src/utils/getFilteredPeople.ts
```diff
@@ -0,0 +1,48 @@
...
+  query: string | null,
+  sex: string | null,
+  centuries: string[],
+) => {
+  let filteredPeople = [...people];
+
+  if (query) {
+    filteredPeople = filteredPeople.filter(person => {
+      return person.name.toLowerCase()
+        .includes(query.toLowerCase())
```
Anna-Tkachenko, line 14: let's prepare the `query` before calling `filter` method
for now you transform it each time on each callback call

Anna-Tkachenko (changes requested):
Check how search params work when user clicks on other user: for now they don't save in url string, let's fix it

Anna-Tkachenko (approved):

PR #187, opened 2023-02-24
First Commit

SalehAkaJim:
[DEMO LINK](https://SalehBUD.github.io/react_people-table-advanced/)

Screenshot 2023-02-28 at 08 25 50


meforoma (changes requested):
test reject

src/components/PeopleFilters.tsx
```diff
@@ -1,89 +1,150 @@
...
+import { useLocation } from 'react-router-dom';
+import { getSearchWith } from '../utils/searchHelper';
+
+type Props = {
+  urlSlug: string | undefined;
+  searchParams: URLSearchParams;
+  setSearchParams: (value: string) => void;
+};
+
+const centuriesList: string[] = ['16', '17', '18', '19', '20'];
```
meforoma, line 12: not great idea to hardcode centuries - what if people db will be supplied with people from other centuries?

construct the centuries array - i'd say it's better to construct it a level where you have the people array and then pass it towards this component.

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,206 @@
...
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const { slug } = useParams();
+
+  // Here it'll take the response from API and add father and mother to
+  // each person, if they exist...
+  const addMotherAndFather = (response: Person[]): Person[] => {
+    const peopleList = response.map((person) => {
+      const findMother = person.motherName
+        ? response.filter((p) => p.name === person.motherName)[0]
```
meforoma, line 32: nonono )
```suggestion
        ? response.find((p) => p.name === person.motherName)
```

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,206 @@
...
+        : null;
+
+      const findFather = person.fatherName
+        ? response.filter((p) => p.name === person.fatherName)[0]
```
meforoma, line 36: use `.find`

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,206 @@
...
+        : null;
+
+      return {
+        ...person,
+        ...(findFather && { father: findFather }),
+        ...(findMother && { mother: findMother }),
```
meforoma, line 42: guess it's ok, yet you could normalise the whole dataset so that props `father` & `mother` are on all records, just some of those will be `null`
```suggestion
        ...{ father: findFather },
        ...{ mother: findMother },
```

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,206 @@
...
+    const orderBy = params.get('order');
+    const sortByText = sortBy === 'name' || sortBy === 'sex';
+
+    if (sortBy && orderBy === 'desc') {
+      setFilteredList((current) => current?.sort((a: Person, b: Person) => {
+        if (sortByText) {
+          return a[sortBy].localeCompare(b[sortBy]);
+        }
+
+        if (sortBy === 'born' || sortBy === 'died') {
```
meforoma, line 97: i'd keep it consistent
```suggestion
    const sortByText = sortBy === 'name' || sortBy === 'sex';
    const sortByInteger = sortBy === 'born' || sortBy === 'died';

    if (sortBy && orderBy === 'desc') {
      setFilteredList((current) => current?.sort((a: Person, b: Person) => {
        if (sortByText) {
          return a[sortBy].localeCompare(b[sortBy]);
        }

        if (sortByInteger) {
```

meforoma: OR even shorter - you already evaluate that sortBy is not falsy, so you can just inverse `sortByText`

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,206 @@
...
 
+import { Person } from '../types';
+import { getPeople } from '../api';
+
+type PageState =
+  | 'loading'
+  | 'error'
+  | 'nothingFound'
+  | 'changeFilter'
+  | 'showTable';
```
meforoma, line 16: it'll work, but it's not a common practice
enum is THE common approach for such

```suggestion
enum PageState {
  Loading,
  // or `Loading = 'loading'` if you rely on text value, otherwise enum items are indexed
  ...;
```

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,206 @@
...
+  const [urlSlug, setUrlSlug] = useState<string>();
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const { slug } = useParams();
+
+  // Here it'll take the response from API and add father and mother to
+  // each person, if they exist...
+  const addMotherAndFather = (response: Person[]): Person[] => {
+    const peopleList = response.map((person) => {
+      const findMother = person.motherName
```
meforoma, line 31: variable name makes me think it's a function - verbs like `find` are used for functions, usually.
```suggestion
      const personMother = person.motherName
```

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,206 @@
...
+        ? response.filter((p) => p.name === person.motherName)[0]
+        : null;
+
+      const findFather = person.fatherName
```
meforoma, line 35: misleading, don't use verbs for non-functions.

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,206 @@
...
+      }));
+    }
+
+    if (sortBy && !orderBy) {
+      setFilteredList((current) => current?.sort((a: Person, b: Person) => {
+        if (sortByText) {
+          return b[sortBy].localeCompare(a[sortBy]);
+        }
+
+        if (sortBy === 'born' || sortBy === 'died') {
```
meforoma, line 111: and here you could reuse the `const sortByInteger`
```suggestion
        if (sortByInteger) {
```

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,206 @@
...
+  }, [people]);
+
+  useEffect(() => {
+    setUrlSlug(slug);
+  }, [slug]);
+
+  useEffect(() => {
+    filter(searchParams);
+    sort(searchParams);
+  }, [searchParams]);
```
meforoma, line 150: why not join the two? 🤔 
```suggestion
  useEffect(() => {
    filter(searchParams);
    sort(searchParams);
  }, [people, searchParams]);

  useEffect(() => {
    setUrlSlug(slug);
  }, [slug]);
```

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,206 @@
...
+              />
+            </div>
+          )}
 
           <div className="column">
             <div className="box table-container">
-              <Loader />
+              {(() => {
+                switch (pageState) {
+                  case 'loading':
```
meforoma, line 172: use enum

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,206 @@
...
   return (
     <>
       <h1 className="title">People Page</h1>
 
       <div className="block">
         <div className="columns is-desktop is-flex-direction-row-reverse">
-          <div className="column is-7-tablet is-narrow-desktop">
-            <PeopleFilters />
-          </div>
+          {pageState === 'showTable' && (
```
meforoma, line 158: use enum

meforoma (changes requested):
@SalehBUD good job 👏🏼 
have a look at my suggestions - i hope you'll agree with the majority.

SalehBUD:
> @SalehBUD good job 👏🏼 have a look at my suggestions - i hope you'll agree with the majority.

Awesome, I'll fix them tonight. Thank you for your time and suggestions. I really appreciate it... 😀🌹

src/components/PeoplePage.tsx
```diff
@@ -117,6 +118,16 @@ export const PeoplePage = () => {
...
   };
 
+  const getAllCenturies = (peopleArray: Person[]) => {
+    const centuries: string[] = [];
+
+    peopleArray.map((person) => {
+      return centuries.push(String(Math.ceil(person.born / 100)));
+    });
+
+    return [...new Set(centuries)];
```
meforoma, line 128: mm, i'd use `.forEach` instead of `.map` :)
just from the method purpose perspective:
- `.map` returns a new array,
- while `.forEach` is intended for iteration

```suggestion
peopleArray.forEach(({ born }) => {
  const bornCentury = Math.ceil(born / 100);
  if (!centuries.includes(bornCentury) {
    centuries.push(bornCentury)
  }

  return centuries
}
```

meforoma (approved):
nice-o 👏🏼

PR #197, opened 2023-02-28
Develop

OlhaYevstifieieva:
[DEMO LINK](https://OlhaYevstifieieva.github.io/react_people-table-advanced/)


src/components/NavBar/Navbar.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { memo } from 'react';
```
GarryRocksman, line 1: If you destructure "memo" you can do in for "FC" also

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,188 @@
...
+    if (sort === sortBy && !isReversed) {
+      return 'fas fa-sort-up';
+    }
+
+    if (sort === sortBy && isReversed) {
+      return 'fas fa-sort-down';
+    }
+
+    return 'fas fa-sort';
+  };
```
GarryRocksman, line 77: nice solution

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,188 @@
...
+            born,
+            died,
+            fatherName,
+            motherName,
+            slug,
+          } = person;
+
+          const hasSelected = slug === personSlug;
+          const selectedMother = people.find(mom => mom.name === motherName);
+          const selectedFather = people.find(dad => dad.name === fatherName);
```
GarryRocksman, line 155: it's better to prepare people's data before rendering because now you use the find method each time when the component is rendered. but you can do it once when data loaded from the server. So you can moove this logic to the halper.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,188 @@
...
+                <PersonLink person={person} />
+              </td>
+
+              <td>{sex}</td>
+              <td>{born}</td>
+              <td>{died}</td>
+
+              <td>
+                {!selectedMother
+                  ? (motherName || '-')
```
GarryRocksman, line 173: you can also use a variable for that

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  return (
+    <NavLink
+      to={`../${slug}`}
+      className={() => (
+        classNames({ 'has-text-danger': sex === 'f' })
+      )}
+    >
+      {name}
+    </NavLink>
```
GarryRocksman, line 21: if i remember right we use NavLink only for navigation links, here you should use Links and class names for danger class

GarryRocksman (changes requested):
Good Job 🔥 rreally clean code, I need to learn from you. But i suppose need some little fixes.

src/utils/sortedPeople.ts
```diff
@@ -0,0 +1,30 @@
...
+  people: Person[],
+  sort: keyof Person | null,
+  isReversed: boolean,
+) => {
+  let sortedPeople = [...people];
+
+  if (sort) {
+    sortedPeople = sortedPeople.sort((a, b) => {
+      switch (sort) {
+        case 'name':
```
ArturErmolenko, line 13: use enum value here

ArturErmolenko (approved):
Well done 🔥

PR #240, opened 2023-03-25
solution

bohdan-gavrilyk:
https://bohdan-gavrilyk.github.io/react_people-table-advanced/


src/components/PeoplePage.tsx
```diff
@@ -1,35 +1,124 @@
...
+        const peopleFromServer = result.map(pep => ({ ...pep }));
+
+        peopleFromServer.forEach(person => {
+          Object.assign(person, {
+            mother: peopleFromServer.find(m => m.name === person.motherName)
+                || null,
+            father: peopleFromServer.find(f => f.name === person.fatherName)
+                || null,
+          });
+        });
```
VitaliyBondarenko1982, line 28:
```suggestion
        const peopleFromServer = result.map(person => ({
          ...person,
          mother: peopleFromServer.find(m => m.name === person.motherName)
                || null,
            father: peopleFromServer.find(f => f.name === person.fatherName)
                || null
        }));
```
we can simplify and avoid additional iteration

src/components/PeopleFilters.tsx
```diff
@@ -16,6 +43,12 @@ export const PeopleFilters = () => {
             type="search"
             className="input"
             placeholder="Search"
+            value={query}
+            onChange={(e) => setsearchParams(
+              getSearchWith(
+                searchParams, { query: e.target.value || null },
+              ),
+            )}
```
VitaliyBondarenko1982, line 51: instead anonymous arrow function better to make function in component and pass reference here

VitaliyBondarenko1982 (changes requested):
good job.
suggest to fix a couple of minor comments

vlad-demchuk (approved):

PR #247, opened 2023-04-07
Develop

halynarydasova:
[DEMO LINK](https://halynarydasova.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,12 +1,46 @@
-export const PeopleFilters = () => {
+import cn from 'classnames';
+import { useState } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import { SearchLink } from './SearchLink';
+
+const allCenturies = ['16', '17', '18', '19', '20'];
+
+export const PeopleFilters:React.FC = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const currentGender = searchParams.get('sex');
+  const pickedCenturies = searchParams.getAll('centuries');
+  const [query, setQuery] = useState('');
+
   return (
     <nav className="panel">
       <p className="panel-heading">Filters</p>
-
       <p className="panel-tabs" data-cy="SexFilter">
-        <a className="is-active" href="#/people">All</a>
-        <a className="" href="#/people?sex=m">Male</a>
-        <a className="" href="#/people?sex=f">Female</a>
+        <SearchLink
+          className={
+            cn({ 'is-active': !currentGender })
+          }
+          params={{ sex: null }}
+        >
+          All
+        </SearchLink>
+        <SearchLink
+          className={
+            cn({
+              'is-active': currentGender === 'm',
+            })
+          }
+          params={{ sex: 'm' }}
+        >
+          Male
+        </SearchLink>
+        <SearchLink
+          className={
+            cn({ 'is-active': currentGender === 'f' })
+          }
+          params={{ sex: 'f' }}
+        >
+          Female
+        </SearchLink>
       </p>
 
       <div className="panel-block">
@@ -15,6 +49,17 @@ export const PeopleFilters = () => {
             data-cy="NameFilter"
             type="search"
             className="input"
+            value={query}
+            onChange={(e) => {
+              setQuery(e.target.value);
+              if (e.target.value.trim() === '') {
+                searchParams.delete('query');
+              } else {
+                searchParams.set('query', e.target.value.trim());
+              }
+
+              setSearchParams(searchParams);
+            }}
             placeholder="Search"
           />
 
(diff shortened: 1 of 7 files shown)


src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,46 @@
...
-        <a className="" href="#/people?sex=f">Female</a>
+        <SearchLink
+          className={
+            cn({ 'is-active': !currentGender })
+          }
+          params={{ sex: null }}
+        >
+          All
+        </SearchLink>
+        <SearchLink
```
Anastasiia-Svintsova, line 26:
```suggestion
        </SearchLink>
        
        <SearchLink
```

src/components/PeopleFilters.tsx
```diff
@@ -15,6 +49,17 @@ export const PeopleFilters = () => {
...
+            onChange={(e) => {
+              setQuery(e.target.value);
+              if (e.target.value.trim() === '') {
+                searchParams.delete('query');
+              } else {
+                searchParams.set('query', e.target.value.trim());
+              }
+
+              setSearchParams(searchParams);
+            }}
```
Anastasiia-Svintsova, line 62: move it to a function

src/components/PeopleFilters.tsx
```diff
@@ -15,6 +49,17 @@ export const PeopleFilters = () => {
...
             type="search"
             className="input"
+            value={query}
+            onChange={(e) => {
+              setQuery(e.target.value);
+              if (e.target.value.trim() === '') {
+                searchParams.delete('query');
+              } else {
+                searchParams.set('query', e.target.value.trim());
+              }
```
Anastasiia-Svintsova, line 59: You can use a ternary operator instead

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,130 @@
...
+    const filteredByCentury = (pickedCenturies.length > 0)
+      ? filteredArray.filter(person => (
+        pickedCenturies.includes(Math.ceil(person.born / 100).toString())))
+      : filteredArray;
+
+    if (query !== '' && query !== null) {
+      const normalizedQuery = query.toLowerCase();
+
+      return filteredByCentury.filter(person => (
+        person.name.toLowerCase().includes(normalizedQuery)
```
Anastasiia-Svintsova, line 45: Create a method to not repeat this part (`.toLowerCase().includes(normalizedQuery)`) 3 times

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,130 @@
...
+    }
+
+    return filteredByCentury;
+  }, [currentGender, pickedCenturies, people, query]);
+
+  const sortedPeople = useMemo(() => {
+    if (sortName) {
+      return [...visiblePeople].sort((a, b) => {
+        const curr = a[sortName as keyof typeof a];
+        const next = b[sortName as keyof typeof b];
```
Anastasiia-Svintsova, line 57: Doesn't it work without it? Just curious 🤔 
```suggestion
        const curr = a[sortName];
        const next = b[sortName];
```

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,130 @@
...
+
+        if (typeof curr === 'string' && typeof next === 'string') {
```
Anastasiia-Svintsova, line 59: Can't you use switch/case and check just a sortName type?

Anastasiia-Svintsova (changes requested):
- Check styles here. Seems that there is a table inside the table 
image

- The reset button is not working

https://user-images.githubusercontent.com/99403198/230711358-aec27686-1315-4344-ac4b-303ce94d7bbc.mov

src/components/PeoplePage.tsx
halynarydasova: it worked, but i have error there and can't deploy

src/components/PeopleFilters.tsx
halynarydasova: Can you explain how i can do that

src/components/PeoplePage.tsx
halynarydasova: But sortName always a string, it's just a key. Value can have different type. Or not?

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,115 @@
-export const PeopleTable = () => {
```
polosanya, line 1: You are missing the import statement for React component here. Add `import React from 'react';` at the beginning.

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,137 @@
...
+
+  useEffect(() => {
+    setIsLoading(true);
+    getPeople().then(result => {
+      setPeople(result);
+    })
+      .catch(() => setIsError(true))
+      .finally(() => setIsLoading(false));
+  }, []);
+  const containsIgnoreCase = (text: string, queryText: string) => (
```
polosanya, line 31: The `containsIgnoreCase` function should be moved outside of the component, as it does not depend on the component's state or props. This will prevent unnecessary re-creation of the function on each render.

polosanya (approved):
Good job!

PR #248, opened 2023-04-09
Solution

lilia-obushenko:
[DEMO LINK](https://lilia-obushenko.github.io/react_people-table-advanced/)


timoxa2t (approved):
Good work

src/components/PersonInfo/PersonInfo.tsx
```diff
@@ -0,0 +1,56 @@
...
+    born,
+    died,
+    motherName,
+    fatherName,
+    mother,
+    father,
+    slug,
+  } = person;
+
+  const { userSlug } = useParams();
```
Oleksandr-kopaevich, line 24: The useParams hook is missing its type. Consider specifying the type for the useParams hook like this: `useParams<{ userSlug: string }>()`

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [hasError, setHasError] = useState(false);
+
+  const [searchParams] = useSearchParams();
+  const sex = searchParams.get('sex') || '';
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries') || [];
```
Oleksandr-kopaevich, line 21: You can replace `searchParams.getAll('centuries') || []` with `searchParams.getAll('centuries')` since `getAll` method returns an array (it can be empty but never `null` or `undefined`).

src/App.tsx
```diff
@@ -1,7 +1,10 @@
-import { PeoplePage } from './components/PeoplePage';
-import { Navbar } from './components/Navbar';
+import { Navigate, Route, Routes } from 'react-router-dom';
+import { PeoplePage } from './pages/PeoplePage/PeoplePage';
+import { Navbar } from './components/Navbar/Navbar';
 
 import './App.scss';
+import { HomePage } from './pages/HomePage/HomePage';
+import { NotFoundPage } from './pages/NotFoundPage/NotFoundPage';
```
Oleksandr-kopaevich, line 7: There are tools for the automatic sorting of imports like extensions or default ones that can be included in your IDE. You can configure hotkeys, which not only sorts imports but also removes unused ones, and merges different lines with the same source into one line. Here is an example of what I'm using. The more helpful tools and hotkeys you'll learn the more productive and useful you can be for your future company.
image

src/App.tsx
```diff
@@ -10,9 +13,15 @@ export const App = () => {
 
       <div className="section">
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
```
Oleksandr-kopaevich, line 16: It's better to have some kind of variable that stores all route values in one place, enum, or just a const. It'll be very clear to see all existing routes in one place and also when in some other components you have to use a redirect or link you can import that variable and avoid using hardcoded values which usually cause bugs and makes refactoring harder.

src/components/SortHeader/SortHeader.tsx
```diff
@@ -0,0 +1,21 @@
+import { SortType } from '../../types/SortType';
+import { SortLink } from '../SortLink/SortLink';
+
+export const SortHeader = () => {
+  return (
+    <tr>
+      {Object.values(SortType).map(value => (
+        <th key={value}>
+          <span className="is-flex is-flex-wrap-nowrap">
+            {`${value.slice(0, 1).toUpperCase()}${value.slice(1)}`}
```
Oleksandr-kopaevich, line 10: It's better to do such transformation using a css rule `text-transform: capitalize;`. But this css will make only the first letter capitalized, so it will not work if the text is uppercased, but in your example, it'll be fine.

src/helpers/helpers.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+export const getFilteredPeople = (
+  people: Person[],
+  sex: string,
+  query: string,
+  centuries: string[],
+  sortBy: string,
+  order: string,
+) => {
```
Oleksandr-kopaevich, line 19: When you have more than 3 arguments for the function it's better to pass them as an object. By this, you don't need to care about the order of the fields, and in your example, it's really hard with 6 fields.
```suggestion
type Props = {
  people: Person[],
  sex: string,
  query: string,
  centuries: string[],
  sortBy: string,
  order: string,
}

export const getFilteredPeople = ({
  people,
  sex,
  query,
  centuries,
  sortBy,
  order,
}) => {
```

src/helpers/helpers.ts
```diff
@@ -0,0 +1,66 @@
...
+          return currPerson[sortBy] - nextPerson[sortBy];
+
+        default:
+          return 0;
+      }
+    });
+  }
+
+  if (order) {
+    visiblePeople.reverse();
```
Oleksandr-kopaevich, line 62: It works right now as expected, but I'd change it to the `sort` method that depends on the value of the `order` variable.

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const filteredPeople = useMemo(() => (
+    getFilteredPeople(
+      people,
+      sex,
+      query,
+      centuries,
+      sortBy,
+      order,
+    )
+  ), [sex, query, centuries, sortBy, order]);
```
Oleksandr-kopaevich, line 35: Seems like you are missing here one more dependency which is `people`

Oleksandr-kopaevich (approved):
Hi @lilia-obushenko, amazing job, 100% from the checklist and really love your style of writing and structuring code. Keep it on a such high level.

PR #251, opened 2023-04-15
Solution

timoxa2t:
[DEMO LINK](https://timoxa2t.github.io/react_people-table-advanced/)


src/App.tsx
```diff
@@ -10,9 +13,17 @@ export const App = () => {
 
       <div className="section">
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
+            <Route path="/" element={<HomePage />} />
```
Oleksandr-kopaevich, line 17: It's better to have some kind of variable that stores all route values in one place, enum, or just a const. It'll be very clear to see all existing routes in one place and also when in some other components you have to use a redirect or link you can import that variable and avoid using hardcoded values which usually cause bugs and makes refactoring harder.

src/components/PeoplePage.tsx
```diff
@@ -1,35 +1,280 @@
...
+    order: null,
+  };
+}
+
+export const PeoplePage: React.FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const { slug: selectedSlug } = useParams();
+  const [searchParams] = useSearchParams();
```
Oleksandr-kopaevich, line 52:
```suggestion
  const [searchParams] = useSearchParams();
  
```

src/components/PeoplePage.tsx
```diff
@@ -1,35 +1,280 @@
...
-              <Loader />
+              {isLoading && <Loader />}
+
+              {isError && (
+                <p data-cy="peopleLoadingError" className="has-text-danger">
+                  Something went wrong
+                </p>
+              )}
+
+              {!isLoading && !isError && !people.length && (
```
Oleksandr-kopaevich, line 175: Use Boolean() when you can have zero
```suggestion
              {Boolean(!isLoading && !isError && !people.length) && (
```
because React renders 0 as a value.
image

src/components/PeoplePage.tsx
```diff
@@ -1,35 +1,280 @@
...
+
+    try {
+      const result = await getPeople();
+
+      const modifiedPeople = result.map(person => {
+        const {
+          motherName,
+          fatherName,
+        } = person;
+        const modifieedPerson = { ...person };
```
Oleksandr-kopaevich, line 71: A typo
```suggestion
        const modifiedPerson = { ...person };
```

Oleksandr-kopaevich (approved):
Well done, but read more careful the task description. You missed this point
`2. Keep search params when navigating within the People page (when selecting a person or clicking the People link).` Also, there are a few other small comments that it's better to fix.

PR #254, opened 2023-04-17
Solution

IVxVI:
[DEMO LINK](https://ivxvi.github.io/react_people-table-advanced/)


src/components/PeopleTable.tsx
```diff
@@ -1,52 +1,100 @@
...
+  const [searchParams] = useSearchParams();
+  const sort = searchParams.get('sort') || null;
+  const order = searchParams.get('order') || null;
+
+  const getSortType = (sortType: SortType) => (
+    {
+      sort: sort === sortType && order ? null : sortType,
+      order: sort === sortType && !order ? 'desc' : null,
+    }
+  );
```
vlad-demchuk, line 25: You should wrap this object in parenthesis to make it more readable and avoid potential issues with automatic semicolon insertion.

src/components/helpers/filterPeople.ts
```diff
@@ -0,0 +1,38 @@
...
+    case centuries.length > 0:
+      filteredPeople = people.filter(person => {
+        const century = Math.ceil(person.born / 100);
+
+        return centuries.includes(century.toString());
+      });
+      break;
+
+    default:
+      filteredPeople = [...people];
```
vlad-demchuk, line 34: Instead of creating a new array with the spread operator, you can simply return `people` directly.

src/components/helpers/sortPeople.ts
```diff
@@ -0,0 +1,25 @@
+import { Person } from '../../types';
+import { SortType } from '../enums/SortType';
+
+export const sortPeople = (
+  people: Person[],
+  sortType: string | null,
+  order: string | null,
+) => {
```
vlad-demchuk, line 8: You should create a copy of the `people` array before sorting to avoid mutating the original array. You can use the spread operator (`...`) to create a shallow copy: `const sortedPeople = [...people];` 🚀

anstsot (changes requested):
It’s almost done! Let’s fix one bug.
I found a case where filters are broken. Try to set the "Male" or "Female" first and then add other filters.
image

PR #255, opened 2023-04-18
advanced people table using search params was created

rashyd-hasratov:
[DEMO LINK](https://rashyd-hasratov.github.io/react_people-table-advanced/)


src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,71 @@
...
+          className={classNames(
+            { 'is-active': sex === null },
+          )}
+          to={{
+            search: getSearchWith(searchParams, { sex: null }),
+          }}
+        >
+          All
+        </Link>
+
```
sergey-lipikhin, line 47: Clean up

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,71 @@
...
+  const {
+    searchParams,
+    setSearchParams,
+  } = useContext(SearchParamsContext);
+
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries');
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    window.console.log('handle method');
```
vladyslav-mosor, line 22: Remove this

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,71 @@
...
+        <Link
+          className={classNames(
+            { 'is-active': sex === 'f' },
+          )}
+          to={{
+            search: getSearchWith(searchParams, { sex: 'f' }),
+          }}
+        >
+          Female
+        </Link> */}
```
vladyslav-mosor, line 68: Delete these coments

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,71 @@
...
+import { SexFilterLink } from './SexFilterLink';
+import { Sex } from '../types/Sex';
+
+const centuriesFilterLinksNames = ['16', '17', '18', '19', '20'];
+
 export const PeopleFilters = () => {
+  const {
+    searchParams,
+    setSearchParams,
+  } = useContext(SearchParamsContext);
```
vladyslav-mosor, line 16: 🔥

src/components/SearchParamsContext/SearchParamsContext.tsx
```diff
@@ -0,0 +1,36 @@
...
+  searchParams: URLSearchParams,
+  setSearchParams: (
+    nextInit: URLSearchParamsInit,
+    navigateOptions?: {
+      replace?: boolean | undefined;
+      state?: any,
+    } | undefined) => void
+}
+
+export const SearchParamsContext = React.createContext<ContextValue>({
```
sergey-lipikhin, line 14: I have also been thinking about adding context, but I realized that `useSearchParams` gives us the ability to interact with `Search Params` anywhere in our program. So `Search Params` are global by nature, it doesn't make much sense to create a context for them, in my opinion

vladyslav-mosor (approved):

sergey-lipikhin (approved):
The `PeopleTable` component is too big. I would recommend creating a `SortLink` component for the `Link` elements in the header and a `PersonItem` component for the rows. It would also be a good idea to move the sorting and filtering logic to a helper function.

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export const PersonLink: React.FC<Props> = React.memo(({ person }) => {
+  const {
+    name,
+    sex,
+    slug,
+  } = person;
+
+  const resolvedPath = useResolvedPath('../').pathname;
```
Oleksandr-kopaevich, line 18: The `useResolvedPath('../').pathname` might cause navigation issues. Consider using a more specific path or using `useRouteMatch` to get the current path.

src/components/PeopleTable.tsx
```diff
@@ -1,683 +1,106 @@
...
-          <td>f</td>
-          <td>1876</td>
-          <td>1956</td>
-          <td>Sophia van Damme</td>
-          <td>Petrus de Milliano</td>
-        </tr>
+export const PeopleTable: React.FC<Props> = ({
+  people,
+}) => {
+  let visiblePeople = [...people];
```
Oleksandr-kopaevich, line 17: It is not necessary to create a copy of the `people` array. You can just use `people` directly and apply filters on it. 🗃️

src/components/PeopleTable.tsx
```diff
@@ -1,683 +1,106 @@
...
-        <tr data-cy="person">
-          <td>
-            <a href="#/people/pieter-haverbeke-1602">Pieter Haverbeke</a>
-          </td>
-          <td>m</td>
-          <td>1602</td>
-          <td>1642</td>
-          <td>-</td>
-          <td>
-            <a href="#/people/lieven-van-haverbeke-1570">
```
Oleksandr-kopaevich, line 68: The hardcoded data should be removed and replaced with the dynamic data rendering using the provided `visiblePeople` array. 🧹

src/components/PeopleTable.tsx
```diff
@@ -1,683 +1,106 @@
...
+      const currentQueryFilterLowercased = currentQueryFilter.toLowerCase();
+
+      return nameLowercased.includes(currentQueryFilterLowercased);
+    });
+  }
+
+  if (currentCenturiesFilter.length) {
+    visiblePeople = visiblePeople.filter(({ born, died }) => {
+      const bornCentury = Math.ceil(born / 1000);
+      const diedCentury = Math.ceil(died / 100);
```
Oleksandr-kopaevich, line 43: It seems like there's a typo in the `bornCentury` calculation. It should be `Math.ceil(born / 100)` instead of `Math.ceil(born / 1000)` 🧐

src/components/PeopleTable.tsx
```diff
@@ -1,683 +1,106 @@
...
-          </td>
-          <td>
-            <a href="#/people/carolus-haverbeke-1832">Carolus Haverbeke</a>
-          </td>
-        </tr>
-      </tbody>
-    </table>
+  return (
+    visiblePeople.length
+      ? (
```
Oleksandr-kopaevich, line 75: Instead of checking for `visiblePeople.length`, it's better to use a conditional rendering with `visiblePeople.length > 0`. It makes the code more explicit and easier to understand. 🔍

src/App.tsx
```diff
@@ -10,9 +14,15 @@ export const App = () => {
 
       <div className="section">
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
+            <Route path="/" element={<HomePage />} />
```
Oleksandr-kopaevich, line 18: It's better to have some kind of variable that stores all route values in one place, enum, or just a const. It'll be very clear to see all existing routes in one place and also when in some other components you have to use a redirect or link you can import that variable and avoid using hardcoded values which usually cause bugs and makes refactoring harder.

src/components/SearchParamsContext/SearchParamsContext.tsx
Oleksandr-kopaevich: @sergey-lipikhin a good explanation, I agree with you.

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,86 @@
...
+                <Loader />
+              )}
+
+              {errorMessage && (
+                <p data-cy="peopleLoadingError" className="has-text-danger">
+                  {errorMessage}
+                </p>
+              )}
+
+              {isLoaded && !people.length && !errorMessage && (
```
Oleksandr-kopaevich, line 71: Use Boolean() to avoid UI bug when in chain of conditions there is a variable with number type, because react renders a `0` when you expected not to load anything at all.
 ```suggestion
              {Boolean(isLoaded && !people.length && !errorMessage) && (
```

File: src/pages/PeoplePage.tsx [hunk 13]
```diff
@@ -0,0 +1,86 @@
...
+                <p data-cy="noPeopleMessage">
+                  There are no people on the server
+                </p>
+              )}
+
+              {isLoaded && people.length && !errorMessage && (
```
Oleksandr-kopaevich, line 77:
```suggestion
              {Boolean(isLoaded && people.length && !errorMessage) && (
```

Oleksandr-kopaevich (approved):
Hey @rashyd-hasratov, approved, but you have one bug related to the next point:
```
4. NameFilter should update the query search param with the text from the input.
show only people with the name, motherName or fatherName that match the query case insensitive;
```
Currently, the search works only for the column `name`, but it also must include results from the columns `father` and `mother`

image
image

PR #256, opened 2023-04-18
Develop react_people-table-advanced

sergey-lipikhin:
[DEMO LINK](https://sergey-lipikhin.github.io/react_people-table-advanced/)


valeriiayarzhemska (approved):
gj!

src/App.tsx
```diff
@@ -1,18 +1,29 @@
...
-      <Navbar />
+      <MainNav />
 
       <div className="section">
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
+            <Route path="/" element={<h1 className="title">Home Page</h1>} />
```
Oleksandr-kopaevich, line 17: It's better to have some kind of variable that stores all route values in one place, enum, or just a const. It'll be very clear to see all existing routes in one place and also when in some other components you have to use a redirect or link you can import that variable and avoid using hardcoded values which usually cause bugs and makes refactoring harder.

src/api.ts
```diff
@@ -8,8 +8,22 @@ function wait(delay: number) {
...
 }
+
+export async function getPeopleWithParents(): Promise<Person[]> {
+  const people = await getPeople();
+
+  return people.map(person => {
+    const mother = people.find(m => m.name === person.motherName);
+    const father = people.find(f => f.name === person.fatherName);
+
+    return {
```
Oleksandr-kopaevich, line 23: It's a good move here for creating a new object on return, I like this way more, without copying person in the start and modifying its fields later like `person.fieldname = newValue`

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,46 @@
...
+    <table
+      data-cy="peopleTable"
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          {Object.values(SortType).map(sorType => (
+            <th key={sorType}>
+              <span className="is-flex is-flex-wrap-nowrap">
+                {sorType.charAt(0).toUpperCase() + sorType.slice(1)}
```
Oleksandr-kopaevich, line 25: Another good way to do such a transformation it's using a CSS rule `text-transform: capitalize;`. But this CSS will make only the first letter capitalized, so it will not work if the text is uppercased, but in your example, it'll be fine.

Oleksandr-kopaevich (approved):
Hi @sergey-lipikhin, really great job here. All points of the checklists are done correctly. I like how you split, structure and organize the code, when I read it, there is no discomfort, like it was written by me. I even don't know what to add here to improve it.

PR #257, opened 2023-04-18
people table with filtering and sorting implemented with using URLSearchParams

vladyslav-mosor:
[DEMO LINK](https://vladyslav-mosor.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters/PeopleFilters.tsx
+++ b/src/components/PeopleFilters/PeopleFilters.tsx
@@ -1,4 +1,8 @@
-import { ChangeEvent, FC, useCallback } from 'react';
+import {
+  ChangeEvent,
+  FC,
+  useCallback,
+} from 'react';
 import classNames from 'classnames';
 import { URLSearchParams } from 'url';
 import { getSearchWith } from '../../utils/helpers';
@@ -20,9 +24,9 @@ export const PeopleFilters: FC<Props> = ({
   searchParams,
   setSearchParams,
 }) => {
-  const handleChange = useCallback((e: ChangeEvent<HTMLInputElement>) => {
+  const handleChange = useCallback((event: ChangeEvent<HTMLInputElement>) => {
     setSearchParams(
-      getSearchWith(searchParams, { query: e.target.value || null }),
+      getSearchWith(searchParams, { query: event.target.value || null }),
     );
   }, [searchParams]);
 
--- a/src/components/PersonInfo/PersonInfo.tsx
+++ b/src/components/PersonInfo/PersonInfo.tsx
@@ -8,6 +8,10 @@ type Props = {
   person: Person,
 };
 
+type Params = {
+  personSlug?: string,
+};
+
 export const PersonInfo: FC<Props> = ({
   person,
 }) => {
@@ -21,7 +25,7 @@ export const PersonInfo: FC<Props> = ({
     mother,
     father,
   } = person;
-  const { personSlug = '' } = useParams();
+  const { personSlug = '' } = useParams<Params>();
 
   return (
     <tr


src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,127 @@
...
+};
+
+export const PeopleFilters: FC<Props> = ({
+  sex,
+  query,
+  centuries,
+  searchParams,
+  setSearchParams,
+}) => {
+  const centuriesList = ['16', '17', '18', '19', '20'];
```
KuharchukDmytro, line 22: move to constants

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,127 @@
...
+              >
+                {century}
+              </SearchLink>
+            ))}
+          </div>
+
+          <div className="level-right ml-4">
+            <SearchLink
+              data-cy="centuryALL"
+              params={{ centuries: [] }}
```
KuharchukDmytro, line 98:
```suggestion
              params={{ centuries: null }}
```

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+      <div className="panel-block">
+        <SearchLink
+          className={classNames(
+            'button is-link is-fullwidth',
+            { 'is-outlined': sex || query || centuries.length > 0 },
+          )}
+          params={{
+            sex: null,
+            centuries: [],
```
KuharchukDmytro, line 118:
```suggestion
            centuries: null,
```

src/utils/helpers.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  if (query) {
+    filteredPeople = filteredPeople.filter(({
+      name,
+      motherName,
+      fatherName,
+    }) => {
+      const availableResults = (name + motherName + fatherName).toLowerCase();
+
+      return availableResults.includes(query.toLowerCase());
```
KuharchukDmytro, line 54: bad idea, its better to check them separate, because for example if i write last chars of person name and then first chars of persons mother name, it will show me that person, but its inccorect
image

KuharchukDmytro (changes requested):
good job, but there are some issues, fix them please 👍 
also there is a bug with error showing when user cannot get data from server, you need to fix it as well :)
image

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,126 @@
...
+  query,
+  centuries,
+  searchParams,
+  setSearchParams,
+}) => {
+  const handleChange = useCallback((e: ChangeEvent<HTMLInputElement>) => {
+    setSearchParams(
+      getSearchWith(searchParams, { query: e.target.value || null }),
+    );
+  }, [searchParams]);
```
KuharchukDmytro, line 27: its better to use full names
```suggestion
  const handleChange = useCallback((event: ChangeEvent<HTMLInputElement>) => {
    setSearchParams(
      getSearchWith(searchParams, { query: event.target.value || null }),
    );
  }, [searchParams]);
```

KuharchukDmytro (approved):
well done👍

src/App.tsx
```diff
@@ -1,20 +1,29 @@
...
+export const App = () => (
+  <div data-cy="app">
+    <NavBar />
+
+    <main className="section">
+      <div className="container">
+        <Routes>
+          <Route path="/" element={<HomePage />} />
+
+          <Route path="home" element={<Navigate to="/" replace />} />
```
Oleksandr-kopaevich, line 17: It's good practice to have some kind of variable that stores all route values in one place, enum, or just a const. It'll be very clear to see all existing routes in one place and also when in some other components you have to use a redirect or link you can import that variable and avoid using hardcoded values which usually cause bugs and makes refactoring harder.

PR #258, opened 2023-04-18
Develop

valeriiayarzhemska:
[DEMO LINK](https://valeriiayarzhemska.github.io/react_people-table-advanced/)


Oleksandr-kopaevich:
Hey @valeriiayarzhemska, be more focused when reading the points of the checklist, you've missed some of them

2. Keep search params when navigating within the People page (when selecting a person or clicking the People link).
**Note:** when I click people search params are cleared on your demo

3. The sidebar with the filters should appear only when people are loaded.
**Note:** I see sidebar when loading results
image

4. NameFilter should update the query search param with the text from the input.
4.1 show only people with the name, motherName or fatherName that match the query case insensitive;
**Note:** Your search is case sensitive. And it searches only column name ignoring mother and father columns.

Please fix those comments.

src/App.tsx
```diff
@@ -1,18 +1,39 @@
...
       <Navbar />
 
       <div className="section">
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
+            <Route
+              path="/"
```
Oleksandr-kopaevich, line 18: It's better to have some kind of variable that stores all route values in one place, enum, or just a const. It'll be very clear to see all existing routes in one place and also when in some other components you have to use a redirect or link you can import that variable and avoid using hardcoded values which usually cause bugs and makes refactoring harder.

src/components/People/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,51 @@
...
+  query: string | null,
+  centuries: string[],
+};
+
+export const PeopleFilters: FC<Props> = (props) => {
+  const {
+    sex,
+    query,
+    centuries,
+  } = props;
```
Oleksandr-kopaevich, line 18:
```suggestion
export const PeopleFilters: FC<Props> = ({
    sex,
    query,
    centuries,
  }) => {
```

src/helpers/helpers.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+};
+
+export const filterPeople = (
+  people: Person[],
+  sex: string | null,
+  query: string | null,
+  centuries: string[],
+): Person[] => {
+  let peopleFiltered = [...people];
```
Oleksandr-kopaevich, line 37: Method `filter` returns a new array every time, so you don’t need to copy it with spread operator.

src/components/People/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,101 @@
...
+    getPeopleFromServer();
+  }, []);
+
+  const filteredPeople = useMemo(() => {
+    return filterPeople(people, sex, query, centuries);
+  }, [people, sex, query, centuries]);
+
+  const sortedPeople = useMemo(() => {
+    return sortPeople(filteredPeople, sort, order);
+  }, [filteredPeople, sort, order]);
```
Oleksandr-kopaevich, line 50: I want to praise you here for splitting logic. It's really great for future code maintenance and debugging.

src/components/People/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,101 @@
...
+                    <p data-cy="noPeopleMessage">
+                      There are no people on the server
+                    </p>
+                  )}
+
+                  {filteredPeople.length <= 0 && (
+                    <p>
+                      There are no people matching the current search criteria
+                    </p>
+                  )}
```
Oleksandr-kopaevich, line 88: This message isn't totally true
`There are no people on the server`
When there is an error in the request, you don't have people locally, but they may exist server. And if you don't have an error, but the server returns zero results you also show this text. So it means one text for two different cases and how a user can distinguish? (in this case by error) But I woudn't even show this text when there is an error in request

src/components/People/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+  } = props;
+
+  return (
+    <table
+      data-cy="peopleTable"
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          {[...Object.keys(SortType)]
```
Oleksandr-kopaevich, line 28: Spread is redundant here. Object keys already creates a new array.
```suggestion
          {Object.keys(SortType)
```

src/components/People/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+            .filter((type) => type.toLowerCase() !== SortType.NONE)
+            .map((type) => (
+              <th>
+                <span className="is-flex is-flex-wrap-nowrap">
+                  {type.slice(0, 1) + type.slice(1).toLowerCase()}
```
Oleksandr-kopaevich, line 33: Another good way to do such a transformation it's using a CSS rule `text-transform: capitalize;`. But this CSS will make only the first letter capitalized, so it will not work if the text is uppercased, but in your example, it'll be fine.

src/components/People/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+type Props = {
+  people: Person[],
+  personId: string,
+};
+
+export const PeopleTable: FC<Props> = (props) => {
+  const {
+    people,
+    personId,
+  } = props;
```
Oleksandr-kopaevich, line 19:
```suggestion
export const PeopleTable: FC<Props> = ({
    people,
    personId,
  }) => {
```

Oleksandr-kopaevich (changes requested):
@valeriiayarzhemska You have nice style and code-splitting skills. Just fix those checklist points and the work can be approved. Good job!

PR #260, opened 2023-04-19
people-table-advanced was implement

KuharchukDmytro:
https://KuharchukDmytro.github.io/react_people-table-advanced/


src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,80 @@
...
       <h1 className="title">People Page</h1>
 
       <div className="block">
         <div className="columns is-desktop is-flex-direction-row-reverse">
           <div className="column is-7-tablet is-narrow-desktop">
-            <PeopleFilters />
+            <PeopleFilters
+              searchParams={serachParams}
+              onSetSearchParams={setSearchParams}
+              query={query}
```
gretoq, line 51: You don't need to pass parameters through props; instead, you can access them directly inside the component where they are needed.

gretoq (changes requested):
From description 3. The sidebar with the filters should appear only when people are loaded.
Please fix it.
Replace PeoplePage.tsx to pages folder, and constants and helpers to utils folder

src/components/PeoplePage.tsx
KuharchukDmytro: on my opinion better to pass them as props, that way we avoid code duplicating

gretoq (approved):

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,50 @@
...
+import classNames from 'classnames';
+import { URLSearchParamsInit } from 'react-router-dom';
+import { getSearchWith } from '../utils/searchHelper';
+import { centuriesList } from '../utils/constants';
+import { SearchLink } from './SearchLink';
+
+type Props = {
+  searchParams: URLSearchParams;
+  onSetSearchParams: (nextInit: URLSearchParamsInit, navigateOptions?: {
+    replace?: boolean | undefined;
```
qoa11a, line 10:
```suggestion
    replace?: boolean;
```
The question mark implies that it could be `undefined`.

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,86 @@
...
+        person => getPersonParents(person, peopleFromServer),
+      );
+
+      setPeople(peopleWithParents);
+    } catch {
+      setHasError(true);
+    } finally {
+      setIsLoading(false);
+    }
+  };
```
qoa11a, line 33: It's better to move the creation of this function inside the `useEffect` where you use it.

qoa11a (approved):
Good job! 👍

PR #262, opened 2023-04-19
Solution

yao157:
[DEMO LINK](https://yao157.github.io/react_people-table-advanced/)


src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,125 @@
+import classNames from 'classnames';
+import { useSearchParams } from 'react-router-dom';
+import { getSearchWith } from '../../utils/searchHelper';
+import { SearchLink } from '../SearchLink';
+
+export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const query = searchParams.get('query') || '';
+  const sex = searchParams.get('sex');
+  const centuries = searchParams.getAll('centuries') || [];
```
vlad-demchuk, line 10: Replace `const centuries = searchParams.getAll('centuries') || [];` with `const centuries = searchParams.getAll('centuries');` because `getAll` method always returns an array, even if it's empty. 🧹

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [searchParams] = useSearchParams();
+  const sort = searchParams.get('sort');
+  const order = searchParams.get('order');
+
+  const loadPeople = async () => {
+    setIsLoading(true);
+    setIsInitialized(false);
+    setIsError(false);
+    try {
+      const serverResp = await getPeople();
```
vlad-demchuk, line 24: Rename the variable `serverResp` to `serverResponse` to improve code readability 📚.

src/utils/helpers.ts
```diff
@@ -0,0 +1,51 @@
+import { Person } from '../types';
+
+export const sortPeople = (
+  peopleArr: Person[],
+  sort: string | null,
+  order: string | null,
```
vlad-demchuk, line 6: Consider using an `options` parameter to group the parameters together in the `sortPeople` function, since it has more than 3 parameters. Example: `sortPeople(peopleArr: Person[], options: { sort: string | null, order: string | null })`

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,114 @@
...
+            sex,
+            motherName,
+            fatherName,
+            born,
+            died,
+            mother,
+            father,
+          } = person;
+          const ifNoMotherLink = motherName || '-';
+          const ifNoFaherLink = fatherName || '-';
```
vlad-demchuk, line 81: There is a typo in the variable name `ifNoFaherLink`, it should be `ifNoFatherLink`. Please correct the variable name.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const query = searchParams.get('query');
+  const centuries = searchParams.getAll('centuries');
+
+  const visiblePeople = useMemo(() => (
+    people.filter(person => {
+      const isSexFilter = sexParam
+        ? person.sex === sexParam
+        : true;
+      const isQueryFilter = query
+        ? person.name.toLowerCase().includes(query)
```
vlad-demchuk, line 25: It's a good practice to also convert the `query` to lowercase when comparing strings, to make the search case-insensitive. Instead of `query`, use `query.toLowerCase()`.

src/App.tsx
```diff
@@ -1,7 +1,9 @@
-import { PeoplePage } from './components/PeoplePage';
-import { Navbar } from './components/Navbar';
-
 import './App.scss';
+import { Navigate, Route, Routes } from 'react-router-dom';
+import { HomePage } from './pages/HomePage';
+import { PageNotFound } from './pages/PageNotFound';
+import { PeoplePage } from './pages/PeoplePage';
+import { Navbar } from './components/Navbar';
```
vlad-demchuk, line 6: Move the `import { Navbar } from './components/Navbar';` statement to the top of the import block to maintain proper import order.

vlad-demchuk (changes requested):

PR #263, opened 2023-04-20
react_people-table-advanced solution

stanislavkulakovskyi:
[DEMO LINK](https://stanislavkulakovskyi.github.io/react_people-table-advanced/)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+export const PeopleTable: React.FC<Props> = React.memo(({
+  people,
+  personSlug,
+}) => {
+  const [searchParams] = useSearchParams();
+  const sort = searchParams.get('sort') as keyof Person;
+  const isReversed = searchParams.get('order') === 'desc';
+  const query = searchParams.get('query');
+  const sexFilter = searchParams.get('sex');
```
andriibezkorovainyi, line 23: It's more likely to name this way 
```suggestion
  const sex = searchParams.get('sex') || '';
```

andriibezkorovainyi (approved):
GJ, I've found some noncritical details that you should correct on my opinion.

src/App.tsx
```diff
@@ -1,20 +1,27 @@
...
+    <Navbar />
 
-export const App = () => {
-  return (
-    <div data-cy="app">
-      <Navbar />
+    <div className="section">
+      <div className="container">
+        <Routes>
+          <Route path="*" element={<NotFoundPage />} />
```
Oleksandr-kopaevich, line 15: It's good practice to have some kind of variable that stores all route values in one place, enum, or just a const. It'll be very clear to see all existing routes in one place and also when in some other components you have to use a redirect or link you can import that variable and avoid using hardcoded values which usually cause bugs and makes refactoring harder.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,125 @@
...
+            fatherName,
+            motherName,
+            slug,
+          } = person;
+
+          const isSelected = slug === personSlug;
+          const mother = people.find(mom => mom.name === motherName);
+          const father = people.find(dad => dad.name === fatherName);
+          const motherChoose = (motherName || '-');
+          const fatherChoose = (fatherName || '-');
```
Oleksandr-kopaevich, line 92: I don't get this part in the variable name `choose`. It's a verb, and verbs are usually used to name functions. You set someone's name as a value, but the variable name doesn't reflect it and it's confusing.

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { Loader } from '../components/Loader';
+import { PeopleFilters } from '../components/PeopleFilters';
+import { PeopleTable } from '../components/PeopleTable';
+import { Person } from '../types';
+
+export const PeoplePage: React.FC = () => {
+  const { personSlug = '' } = useParams();
+
+  const [people, setPeople] = useState<Person[]>([]);
+  const [hasError, setHasError] = useState(false);
```
Oleksandr-kopaevich, line 13: It's more comfortable to have `errorMessage` state and save there a string error. If it's empty then there are no errors, if it has value then it can be used for any error.

src/utils/getFilteredPeople.ts
```diff
@@ -0,0 +1,50 @@
+import { Person } from '../types';
+
+export const getFilteredPeople = (
+  people: Person[],
+  query: string | null,
+  sex: string | null,
+  centuries: string[],
+) => {
+  let filteredPeople = [...people];
```
Oleksandr-kopaevich, line 9: Method `filter` returns a new array every time, so you don’t need to copy it with the spread operator.

Oleksandr-kopaevich (approved):
Hey @stanislavkulakovskyi, good job. All checklist points are done and there are no critical moments in the code, just small minor suggestions for improvement. Approved!

PR #264, opened 2023-04-20
people table with links

novak-k:
[DEMO LINK](https://novak-k.github.io/react_people-table-advanced/)


src/components/Navbar.tsx
```diff
@@ -8,15 +11,22 @@ export const Navbar = () => {
...
+            to="/"
+            className={({ isActive }) => classNames('navbar-item',
+              { 'has-background-grey-lighter': isActive })}
+          >
+            Home
+          </NavLink>
 
-          <a
+          <NavLink
+            to="people"
```
TarasHoliuk, line 23: It's better to use an absolute path for the NavLink `to` property. Add a leading slash to make it `/people` instead of `people`. 🛣️

src/components/PeoplePage.tsx
```diff
@@ -1,8 +1,77 @@
-import { PeopleFilters } from './PeopleFilters';
+import { useEffect, useMemo, useState } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import { PeopleFilters } from './PeopleFilters/PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { Person } from '../types';
+import { getPeople } from '../api';
+import { FilteredPeople } from './FilteredPeople';
```
TarasHoliuk, line 8: Consider renaming 'FilteredPeople' to 'filterPeople' to indicate it's a function, not a component.

src/components/FilteredPeople.tsx
```diff
@@ -0,0 +1,55 @@
+import { SortType } from '../types/SortType';
+import { Person } from '../types/Person';
+
+export const FilteredPeople = (
```
TarasHoliuk, line 4: FilteredPeople function has too many parameters, consider using an `options` parameter to group them together.

src/components/Links/SortLink.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { FilterType } from '../../types/FilterType';
+
+type Props = {
+  sortFilter: string,
+};
+
+export const SortLink: React.FC<Props> = ({ sortFilter }) => {
+  const [searchParams] = useSearchParams();
+  const sort = searchParams.get(FilterType.SORT) || null;
+  const order = searchParams.get(FilterType.ORDER);
```
TarasHoliuk, line 13: It is a good practice to provide a default value for the 'order' variable in case it is not found in the searchParams. You can add '|| null' or default value you want to be set, at the end of the line, like you did for the 'sort' variable.

src/App.tsx
```diff
@@ -10,9 +10,20 @@ export const App = () => {
...
       <div className="section">
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
+            <Route path="/" element={<h1 className="title">Home Page</h1>} />
+            <Route path="home" element={<Navigate to="/" replace />} />
+
+            <Route path="people">
```
TarasHoliuk, line 17: It would be better to use the `index` prop on the main `Route` for the 'people' path instead of creating a nested `Route` with the `index` prop.

src/components/FilteredPeople.tsx
TarasHoliuk: also, use camelCase for not component functions and add a verb that explains what the function does, i.e `getFilteredPeople` or `filterPeople`

TarasHoliuk (changes requested):
GJ! Added a few minor comments

src/components/Navbar.tsx
novak-k: fixed

src/components/FilteredPeople.tsx
novak-k: fixed

src/components/Links/SortLink.tsx
novak-k: added

src/App.tsx
novak-k: not sure that I get it correct, but changed

tetiana-ivanchenko (approved):

PR #266, opened 2023-04-23
react peopla table advanced

andriibezkorovainyi:
[DEMO LINK](https://andriibezkorovainyi.github.io/react_people-table-advanced/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
+import {Navigate, Route, Routes} from "react-router-dom";
 import { PeoplePage } from './components/PeoplePage';
-import { Navbar } from './components/Navbar';
 
+import { Navbar } from './components/Navbar';
 import './App.scss';
 
 export const App = () => {
@@ -10,9 +11,33 @@ export const App = () => {
 
       <div className="section">
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
+            <Route
+              path="*"
+              element={<h1 className="title">Page not found</h1>}
+            />
+            <Route
+              path="home"
+              element={<Navigate to="/" replace />}
+            />
+
+            <Route
+              path="/"
+              element={<h1 className="title">Home Page</h1>}
+            />
+
+            <Route path="people">
+              <Route
+                index
+                element={<PeoplePage />}
+              />
+
+              <Route
+                path=":personSlug"
+                element={<PeoplePage />}
+              />
+            </Route>
+          </Routes>
         </div>
       </div>
     </div>
--- a/src/components/ParentLink.tsx
+++ b/src/components/ParentLink.tsx
@@ -1,0 +1,20 @@
+import { FC } from 'react';
+import {PersonLink} from "./PersonLink";
+import {Person} from "../types";
+
+interface Props {
+  parentName: string | null,
+  parent: Person | undefined,
+}
+
+export const ParentLink: FC<Props> = ({ parentName, parent }) => {
+  if (!parentName) {
+    return <span>-</span>;
+  }
+
+  if (parent === undefined) {
+    return <span>{parentName}</span>;
+  }
+
+  return <PersonLink person={parent} />;
+};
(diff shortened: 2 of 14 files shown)


src/App.tsx
```diff
@@ -10,9 +11,33 @@ export const App = () => {
...
+              path="home"
+              element={<Navigate to="/" replace />}
+            />
+
+            <Route
+              path="/"
+              element={<h1 className="title">Home Page</h1>}
+            />
+
+            <Route path="people">
```
Oleksandr-kopaevich, line 29: It's good practice to have some kind of variable that stores all route values in one place, enum, or just a const. It'll be very clear to see all existing routes in one place and also when in some other components you have to use a redirect or link you can import that variable and avoid using hardcoded values which usually cause bugs and makes refactoring harder.

src/components/ParentLink.tsx
```diff
@@ -0,0 +1,20 @@
...
+  if (!parentName) {
+    return <span>-</span>;
+  }
+
+  if (parent === undefined) {
+    return <span>{parentName}</span>;
+  }
+
+  return <PersonLink person={parent} />;
+};
```
Oleksandr-kopaevich, line 20: This way/order is more often used
```suggestion
  if (parent) {
    return <PersonLink person={parent} />;
  }

  return <span>{parentName}</span>;
};
```

src/components/PeopleFilters.tsx
```diff
@@ -1,21 +1,74 @@
...
+        value.forEach(part => {
+          searchParams.append(key, part);
+        });
+      } else {
+        searchParams.set(key, value);
+      }
+    });
+
+    setSearchParams(searchParams);
+  }
```
Oleksandr-kopaevich, line 28: This can be split to a function helper that returns a new searchParams and then use setSearchParams to set it to the state. The component's body should stay as small and clear as possible.

src/components/PeopleFilters.tsx
```diff
@@ -1,21 +1,74 @@
...
+        >
+          Male
+        </SearchLink>
+
+        <SearchLink
+          params={{ sex: 'f' }}
+          className={classNames({ 'is-active': sex === 'f' })}
+        >
+          Female
+        </SearchLink>
```
Oleksandr-kopaevich, line 60: When you see a boilerplate code that duplicates itself then think about how to make it shorter. For example you can use object/enum where text is a key and the value is f/m/''. Then using map you can render any amount of links and have one object as config for it.

src/components/PeoplePage.tsx
```diff
@@ -1,35 +1,93 @@
...
+
+        if (mother) {
+          personWithParents.mother = mother;
+        }
+
+        if (father) {
+          personWithParents.father = father;
+        }
+
+        return personWithParents;
```
Oleksandr-kopaevich, line 39: More often we collect data for a new field in separate variables and then when making return we spread everything into a new object. This way it's easier to read and change.
```suggestion
        const mother = getParentByName(person.motherName, peopleFromServer);
        const father = getParentByName(person.fatherName, peopleFromServer);

        return {
            ...person,
            ...mother && { mother },
            ...father && { father },
        };
```

src/components/PersonInfo.tsx
```diff
@@ -0,0 +1,57 @@
+import { FC } from 'react';
+import classNames from 'classnames';
+import {Person} from "../types";
+import {PersonLink} from "./PersonLink";
+import {ParentLink} from "./ParentLink";
+
+interface Props {
+  person: Person,
+  // people: Person[],
```
Oleksandr-kopaevich, line 9: When submitting a PR check your code and clean it from a temporary code.
```suggestion
```

src/utils/filterPeople.ts
```diff
@@ -0,0 +1,47 @@
...
+  const birthYear = person.died - years;
+  const personCentury = String(Math.floor(birthYear / 100) + 1);
+
+  return centuries.includes(personCentury);
+}
+
+export function filterPeople(
+  people: Person[], filterParams: [string, string, string[]],
+) {
+  let filteringPeople = [...people];
```
Oleksandr-kopaevich, line 23: Method `filter` returns a new array every time, so you don’t need to copy it with the spread operator.

PR #270, opened 2023-04-29
● Solution

d-hubych:
- [DEMO LINK](https://d-hubych.github.io/react_people-table-advanced/)


src/components/PersonLink.tsx
```diff
@@ -0,0 +1,87 @@
...
+    born,
+    died,
+    motherName,
+    mother,
+    fatherName,
+    father,
+    slug,
+  } = person;
+
+  const parent = (
```
etojeDenys, line 29: Consider renaming 'parent' function to be more descriptive, such as 'renderParentLink'.

src/components/Navbar.tsx
```diff
@@ -1,24 +1,22 @@
...
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
-        </div>
+        <a
+          aria-current="page"
+          className="navbar-item has-background-grey-lighter"
+          href="#/people"
```
etojeDenys, line 15: Use React Router's Link component instead of an anchor tag for internal navigation.

src/components/SortLink.tsx
```diff
@@ -0,0 +1,50 @@
...
+  sortBy: SortBy,
+  isSortingNow: boolean,
+};
+
+export const SortLink: FC<Props> = ({ sortBy, isSortingNow }) => {
+  const [searchParams] = useSearchParams();
+  const currentSort = searchParams.get('sort');
+  const currentOrder = searchParams.get('order');
+
+  const handleSortParams = (type: SortBy) => {
```
etojeDenys, line 17: Change 'handleSortParams' to a useCallback to prevent unnecessary re-renders.

src/components/PeopleTable.tsx
```diff
@@ -1,4 +1,78 @@
...
+
+type Props = {
+  people: Person[],
+};
+
+export const PeopleTable: FC<Props> = ({ people }) => {
+  const { slug } = useParams();
+  const [searchParams] = useSearchParams();
+
+  const sex = searchParams.get('sex');
```
etojeDenys, line 16: You can destructure the searchParams object to make the code cleaner. For example, const { sex, centuries, query, sort, order } = searchParams;

src/components/PeopleTable.tsx
```diff
@@ -1,4 +1,78 @@
...
+      )
+    ) {
+      return false;
+    }
+
+    return true;
+  });
+
+  const onSort = () => {
+    switch (sort) {
```
etojeDenys, line 49: Instead of using an arrow function for 'onSort', consider using a regular function to avoid problems with 'this' binding

src/components/PeopleTable.tsx
```diff
@@ -1,4 +1,78 @@
...
+        filteredPeople.sort((person1, person2) => person1.slug
+          .localeCompare(person2.slug));
+    }
+
+    if (order) {
+      filteredPeople.reverse();
+    }
+  };
+
+  if (sort) {
```
etojeDenys, line 72: It is better to call the 'onSort()' function inside a useEffect hook to avoid unnecessary re-renders

etojeDenys (changes requested):
your link does not work

src/components/Navbar.tsx
d-hubych: Thanks, there was a bigger problem here. Navbar was not used. 😅

src/App.tsx
```diff
@@ -1,20 +1,42 @@
-import { PeoplePage } from './components/PeoplePage';
+import { FC } from 'react';
+import {
+  // NavLink,
```
VitaliyBondarenko1982, line 3: Remove commented code

src/App.tsx
```diff
@@ -1,20 +1,42 @@
...
+  Navigate,
+  Route,
+  Routes,
+} from 'react-router-dom';
+// import cn from 'classnames';
```
VitaliyBondarenko1982, line 8: commented code

src/components/PeopleTable.tsx
VitaliyBondarenko1982: not fixed. we can avoid useEffect and instead wrap onSort in useMemo and return sortedPeople. also better to wrap filteredPeople in useMemo
```
  const filteredPeople = useMemo(() => {
    return people.filter(person => {
      if (sex && person.sex !== sex) {
        return false;
      }

      if (centuries.length) {
        const century = Math.ceil(person.born / 100);

        if (!centuries.includes(`${century}`)) {
          return false;
        }
      }

      if (query
        && !(
          person.name.toLowerCase().includes(query)
          || person.motherName?.toLowerCase().includes(query)
          || person.fatherName?.toLowerCase().includes(query)
        )
      ) {
        return false;
      }

      return true;
    });
  }, [people, query, sort, order]);

  ```
    const sortedPeople = useMemo(() => {
      const sorted = [...filteredPeople];

    switch (sort) {
      case SortBy.name:
      case SortBy.sex:
        sorted.sort((person1, person2) => person1[sort]
          .localeCompare(person2[sort]));
        break;

      case SortBy.born:
      case SortBy.died:
        sorted
          .sort((person1, person2) => person1[sort] - person2[sort]);
        break;

      default:
        sorted.sort((person1, person2) => person1.slug
          .localeCompare(person2.slug));
    }

    if (order) {
      sorted.reverse();
    }

      return sorted;
    }, [order, sort, filteredPeople]);

**Review by VitaliyBondarenko1982 — CHANGES_REQUESTED** [review]

File: src/pages/PeoplePage.tsx [hunk 9]
```diff
@@ -0,0 +1,75 @@
...
+import { useMatch } from 'react-router-dom';
+import { Loader } from '../components/Loader';
+import { PeopleTable } from '../components/PeopleTable';
+import { Person } from '../types/Person';
+import { getPeople } from '../api';
+import { PeopleFilters } from '../components/PeopleFilters';
+
+export const PeoplePage: FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoader, setIsLoader] = useState(false);
```
etojeDenys, line 11: Consider renaming 'isLoader' to 'isLoading' for better readability.

src/components/SortLink.tsx
```diff
@@ -0,0 +1,52 @@
...
+  sortBy: SortBy,
+  isSortingNow: boolean,
+};
+
+export const SortLink: FC<Props> = ({ sortBy, isSortingNow }) => {
+  const [searchParams] = useSearchParams();
+  const currentSort = searchParams.get('sort');
+  const currentOrder = searchParams.get('order');
+
+  const handleSortParams = useCallback(
```
etojeDenys, line 17: The 'type' parameter in the handleSortParams function is redundant. You can use the 'sortBy' prop directly inside the function.

etojeDenys (approved):
lgtm

PR #273, opened 2023-05-03
Solution

katya-suvorova:
https://katya-suvorova.github.io/react_people-table-advanced/


src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,149 @@
+import { useEffect, useMemo, useState } from 'react';
+import { useParams, useSearchParams } from 'react-router-dom';
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { Person } from '../types';
+import { getPeople } from '../api';
+import { SortType } from '../types/SortType';
+
+const filterPeople = (
```
anstsot, line 10: it would be great to create separate js files in `utils` folder and place these functions there

anstsot (changes requested):
Good job, just one small fix to add. 
There is an issue that "no data" text and loader are shown at the beginning. 
image

PR #278, opened 2023-05-15
Solution react_people-table-advanced

NazariiAlieksieiev:
[DEMO LINK](https://NazariiAlieksieiev.github.io/react_people-table-advanced/)


src/components/PersonLink.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export type Props = {
+  to: string;
+  text: string;
+  sex?: string;
+};
+
+export const PersonLink: FC<Props> = ({ to, text, sex }) => (
+  <NavLink
+    to={`../${to}`}
```
etojeDenys, line 13: Use the `to` prop directly without adding '../', since it might cause incorrect routing.

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,18 @@
...
+    className={sex === 'f' ? classNames('has-text-danger') : ''}
```
etojeDenys, line 14: You can simplify the className assignment using the ternary operator directly inside the string instead of using 'classNames'.

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,130 @@
...
+                <th key={param}>
+                  <span className="is-flex is-flex-wrap-nowrap">
+                    {param}
+                    <SearchLink
+                      params={{
+                        sort: isSelectedSortType && order
+                          ? null
+                          : param.toLowerCase(),
+                        order: isSelectedSortType && !order
+                          ? 'decs'
```
etojeDenys, line 41: There is a typo in the order value, it should be 'desc' instead of 'decs'.

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,130 @@
...
+import { SearchLink } from './SearchLink';
+
+interface Props {
+  people: Person[]
+  link: string
+  sort: string
+  order: string
+}
+
+export const PeaopleTable: FC<Props> = ({
```
etojeDenys, line 14: There is a typo in the component name, it should be 'PeopleTable' instead of 'PeaopleTable'.

src/components/PeoplePage.tsx
```diff
@@ -1,34 +1,198 @@
-import { PeopleFilters } from './PeopleFilters';
+import {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
+import { useParams, useSearchParams } from 'react-router-dom';
+import { Person } from '../types';
+import { getPeople } from '../api';
+import { PeaopleTable } from './PeopleTable';
```
etojeDenys, line 7: There is a typo in the component name. It should be 'PeopleTable' instead of 'PeaopleTable'.

src/components/PeoplePage.tsx
```diff
@@ -1,34 +1,198 @@
...
+                </p>
+              )}
 
-              <p>There are no people matching the current search criteria</p>
-
-              <PeopleTable />
+              {isInPorcess
+                ? <Loader />
+                : (filterAndSortPeople().length > 0 && !isInPorcess) && (
+                  <PeaopleTable
```
etojeDenys, line 185: There is a typo in the component name. It should be 'PeopleTable' instead of 'PeaopleTable'.

src/components/PeoplePage.tsx
```diff
@@ -1,34 +1,198 @@
...
+        sortedPeople = sortedPeople
+          .sort((
+            personA,
+            personB,
+          ) => (personA[sort] - personB[sort]) * sortDirection);
+        break;
+      default:
+        break;
+    }
+
```
etojeDenys, line 128: better to create some enum and use it here

etojeDenys (changes requested):

src/components/PersonLink.tsx
NazariiAlieksieiev: I don't now why, but it is only one way when its works correctly

src/components/PersonLink.tsx
etojeDenys, line 13: Use a relative path instead of adding '../' to the 'to' prop. Replace with 'to={to}'.

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,51 @@
...
+}
+
+export const PeopleFilters: React.FC<Props> = ({
+  searchParams,
+  centuries,
+  query,
+  sex,
+  onQueryChange,
+}) => {
+  const centuriesArray = ['16', '17', '18', '19', '20'];
```
etojeDenys, line 22: Instead of hardcoding the centuriesArray, it's better to pass it as a prop or import it from a constants file.

src/components/PeopleFilters.tsx
```diff
@@ -27,66 +68,52 @@ export const PeopleFilters = () => {
...
-
-            <a
-              data-cy="century"
-              className="button mr-1"
-              href="#/people?centuries=20"
-            >
-              20
-            </a>
+            {centuriesArray.map(centurie => (
+              <Link
```
etojeDenys, line 72: Use a more meaningful variable name instead of 'centurie'. Consider using 'century' instead.

src/components/PeoplePage.tsx
```diff
@@ -1,34 +1,199 @@
...
+import { Loader } from './Loader';
+import { PeopleFilters } from './PeopleFilters';
+import { getSearchWith } from '../utils/searchHelper';
+import { SortOption } from '../types/SortOption';
 
 export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [error, setError] = useState<boolean>(false);
+
+  const [isInPorcess, setIsInProcess] = useState<boolean>(false);
```
etojeDenys, line 17: Typo in the variable name 'isInPorcess', it should be 'isInProgress'

src/components/PeoplePage.tsx
```diff
@@ -1,34 +1,199 @@
...
+      try {
+        const fetchedPerson = await getPeople();
+
+        setPeople(fetchedPerson);
+      } catch (innerError) {
+        setError(true);
+      } finally {
+        setIsInProcess(false);
+      }
+    };
```
etojeDenys, line 46: Typo in the function name 'fetchPerson', should be 'fetchPeople'

src/components/PeoplePage.tsx
```diff
@@ -1,34 +1,199 @@
...
+        mother,
+      });
+    });
+
+    return result;
+  }, [people]);
+
+  const filtrForPeople = () => {
+    const getCenturie = (date: number) => {
+      return (Math.floor(date / 100) + 1).toString();
```
etojeDenys, line 78: Consider renaming the function 'getCenturie' to 'getCentury'

etojeDenys (approved):

PR #279, opened 2023-05-15
Ready

sergik94:
https://sergik94.github.io/react_people-table-advanced/


src/components/Parent.tsx
```diff
@@ -0,0 +1,17 @@
...
+  parent: Person | undefined;
+  parentName: string | null;
+};
+
+export const Parent: React.FC<Props> = ({ parent, parentName }) => {
+  return (
+    <>
+      {!!parent && <PersonLink person={parent} />}
+      {!!parentName && !parent && `${parentName}`}
+      {!parentName && '-'}
```
OleziO, line 14: Wrap text with span tag (or any text tag)

OleziO (changes requested):

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,111 @@
...
+    }
+  }
+
+  function filterByQuery(str: string | null) {
+    return str?.toLowerCase().includes(query?.toLowerCase() || '') || false;
+  }
+
+  const shownPeopleList = useMemo(() => {
+    let filteredArray = [...people];
+
```
IvanFesenko, line 42: No need to create copy, cuz  filter return new Array `sortPeople` create new one

IvanFesenko (approved):

PR #280, opened 2023-05-18
reused code from 1st part of task

PavloReutskiy:
[DEMO LINK](https://PavloReutskiy.github.io/react_people-table-advanced/)


src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,110 @@
...
+          <span className="icon is-left">
+            <i className="fas fa-search" aria-hidden="true" />
+          </span>
+        </p>
+      </div>
+
+      <div className="panel-block">
+        <div className="level is-flex-grow-1 is-mobile" data-cy="CenturyFilter">
+          <div className="level-left">
+            {['16', '17', '18', '19', '20'].map(century => (
```
tiserett, line 66: I recommend putting this value into separate variable

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,119 @@
...
+export const PeoplePage: FC = () => {
+  const { people, isLoading, errorMessage } = useFetch(API_URL);
+  const [searchParams] = useSearchParams();
+  const gender = searchParams.get('sex') || null;
+  const query = searchParams.get('query') || null;
+  const centuries = searchParams.getAll('centuries') || [];
+  const sort = searchParams.get('sort') || null;
+  const order = searchParams.get('order') || null;
+
+  const getPerent = (parentName: string | null) => {
```
tiserett, line 20: pErent -> pArent

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+      case 'died':
+        return order === 'desc'
+          ? b.died - a.died
+          : a.died - b.died;
+
+      default:
+        return 0;
+    }
+  });
```
tiserett, line 78: better move these functions out of component

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,119 @@
...
+            <div className="box table-container">
+              {isLoading && (<Loader />)}
+
+              {errorMessage && (
+                <p data-cy="peopleLoadingError" className="has-text-danger">
+                  {errorMessage}
+                </p>
+              )}
+
+              {people?.length === 0 && (
```
tiserett, line 100: you can use shorthand people?.length, it should work the same way
consider fixing it here and in other occurencies

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,90 @@
...
+  return (
+    <table
+      data-cy="peopleTable"
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          <th>
+            <TableHeader title="Name" />
+          </th>
```
tiserett, line 25: why don't you put <th> inside of component?

src/components/TableHeader/TableHeader.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const TableHeader: FC<Props> = ({ title }) => {
+  const [searchParams] = useSearchParams();
+  const sort = searchParams.get('sort') || null;
+  const order = searchParams.get('order') || null;
+  const sortValue = title.toLowerCase();
+
+  const getParams = (searchParam: string) => {
+    if (sort === null && order === null) {
```
tiserett, line 17: if (!sort && !order) etc.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,90 @@
...
+          <th>
+            <TableHeader title="Born" />
+          </th>
+
+          <th>
+            <TableHeader title="Died" />
+          </th>
+
+          <th>Mother</th>
+          <th>Father</th>
```
tiserett, line 40: can't you make use of your TableHeader  component?

tiserett (changes requested):
good job, but there are few minor problems

src/components/PeopleTable/PeopleTable.tsx
PavloReutskiy: I can't because TableHeader component add buttons, so I use it only for elements that can be filtered. But these two headers don't have buttons according to example
image

src/components/PeopleTable/PeopleTable.tsx
PavloReutskiy: I'm not sure that I understand correct. I use TableHeader component for every <th> that will work like button

qoa11a (approved):

PR #283, opened 2023-05-22
solution

PashaMalyshkin:
[DEMO LINK](https://PashaMalyshkin.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,12 +1,55 @@
-export const PeopleFilters = () => {
+import { FC } from 'react';
+import cn from 'classnames';
+import { Link, useSearchParams } from 'react-router-dom';
+import { SexCategory } from '../types/SexCategory';
+import { SearchLink } from './SearchLink';
+import { getSearchWith } from '../utils/searchHelper';
+
+export const PeopleFilters: FC = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const sex = searchParams.get('sex') || SexCategory.NONE;
+  const centuries = searchParams.getAll('centuries') || [];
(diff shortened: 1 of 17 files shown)


src/components/PeopleFilters.tsx
```diff
@@ -27,66 +72,46 @@ export const PeopleFilters = () => {
...
         <div className="level is-flex-grow-1 is-mobile" data-cy="CenturyFilter">
           <div className="level-left">
-            <a
-              data-cy="century"
-              className="button mr-1"
-              href="#/people?centuries=16"
-            >
-              16
-            </a>
+            {['16', '17', '18', '19', '20'].map(century => (
```
qoa11a, line 75: It's better to move this array to a variable.

src/components/PeoplePage.tsx
```diff
@@ -1,35 +1,125 @@
...
+  const [people, setPeople] = useState<Person[]>([]);
+  const [loadingError, setLoadingError] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+  const [searchParams] = useSearchParams();
+
+  const sex = searchParams.get('sex') || SexCategory.NONE;
+  const centuries = searchParams.getAll('centuries') || [];
+  const query = searchParams.get('query') || '';
+
+  const loadPeople = useCallback(async () => {
```
qoa11a, line 26: As you use this function only inside the `useEffect` you may move it there. Don't just use `useCallback` without a reason.

image


https://react.dev/reference/react/useCallback

qoa11a (changes requested):
Good job!
Just a few suggestions.

qoa11a (approved):

PR #289, opened 2023-05-24
completed

Zad0rnik:
[DEMO LINK](https://Zad0rnik.github.io/react_people-table-advanced/)


src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [searchParams] = useSearchParams();
+
+  const sex = searchParams.get('sex') || Sex.ALL;
+  const query = searchParams.get('query')?.toLowerCase() || '';
+  const centuries = searchParams.getAll('centuries') || [];
+  const order = searchParams.get('order');
+  const sort = searchParams.get('sort');
+
+  const loadPeople = useCallback(async () => {
```
qoa11a, line 33: As you use this function only inside the useEffect you may move it there. Don't just use useCallback without a reason.

image

https://react.dev/reference/react/useCallback

I'd recommend you read more about the useCallback hook and then fix all the cases when you use it without a reason.

qoa11a (changes requested):
Good job in general! 👍 
I'd recommend reading more about useMemo and useCallback. They are not magical functions that just boost performance. You should use them only on specific occasions about which you can read in React documentation.

qoa11a (approved):

PR #307, opened 2023-06-07
solution

ghostargg:
[DEMO LINK](https://ghostargg.github.io/react_people-table-advanced/)


src/PageNotFound.tsx
```diff
@@ -0,0 +1,5 @@
+export const PageNotFound = () => {
+  return (
```
kasiakubisiak, line 2: you don't need return here

src/components/Home.tsx
```diff
@@ -0,0 +1,5 @@
+export const HomePage = () => {
+  return (
```
kasiakubisiak, line 2: you don't need return here

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,152 @@
...
+  };
+
+  useEffect(() => {
+    loadPeople();
+  }, []);
+
+  const visiblePeople = useMemo(() => {
+    let filteredPeople = people;
+
+    if (sex) {
```
kasiakubisiak, line 51: maybe lets do switch/case for sex, query and centuries?

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,152 @@
...
-              </p>
+              {!isLoading && people.length === 0
+                && (
+                  <p data-cy="noPeopleMessage">
+                    There are no people on the server
+                  </p>
+                )}
 
-              <p>There are no people matching the current search criteria</p>
+              {!isLoading && !hasError && visiblePeople.length === 0
```
kasiakubisiak, line 135:
```suggestion
              {!isLoading && !hasError && !visiblePeople.length
```

kasiakubisiak (changes requested):

src/components/PeoplePage.tsx
ghostargg: those are all separate filters that co-work i dont know how i can implement it with switch case.

PR #311, opened 2023-06-13
solution

AnastasiiaVorobets:
[DEMO LINK](https://AnastasiiaVorobets.github.io/react_people-table-advanced/)


AnastasiiaVorobets:
This is not a completely finished code. I am finishing some more points.

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,95 @@
...
+export const PeoplePage = () => {
+  const [searchParams] = useSearchParams();
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isError, setIsError] = useState(false);
+  const [isLoading, setLoading] = useState(false);
+  const sex = searchParams.get('sex');
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries') || null;
+  const sortType = searchParams.get('sort');
+  const order = searchParams.get('order') || null;
```
volodymyr-soltys97, line 20: Use an empty string instead of null for the default value of 'order' to avoid possible issues when using string methods or checking for truthy values.

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { sortPeople } from '../utils/sortPeople';
+
+export const PeoplePage = () => {
+  const [searchParams] = useSearchParams();
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isError, setIsError] = useState(false);
+  const [isLoading, setLoading] = useState(false);
+  const sex = searchParams.get('sex');
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries') || null;
```
volodymyr-soltys97, line 18: Use an empty array instead of null for the default value of 'centuries' to avoid possible issues when iterating or using array methods.

src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,37 @@
+import classNames from 'classnames';
+import { NavLink } from 'react-router-dom';
+
+export const Navigation = () => (
```
volodymyr-soltys97, line 4: Add the 'import React' statement at the beginning of the file.

volodymyr-soltys97, line 4: The Navigation component should be exported as a default export. Change it to 'export default const Navigation = () => (...)'.

src/utils/sortPeople.ts
```diff
@@ -0,0 +1,37 @@
+import { Person } from '../types/Person';
+
+export const sortPeople = (
+  people: Person[],
+  sortType: string | null,
```
volodymyr-soltys97, line 5: Consider replacing 'string | null' with a more specific type or enum for sortType parameter to avoid passing incorrect values.

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+export const PeopleFilters: FC<PeopleFiltersPerops> = ({
+  query,
+  sex,
+  centuries,
+}) => {
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const centuriesArray: string[] = ['16', '17', '18', '19', '20'];
+
```
volodymyr-soltys97, line 21: You can consider creating a constants file to store the 'centuriesArray' and import it whenever needed, rather than hardcoding it here.

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,131 @@
+import React, { FC } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import classNames from 'classnames';
+import { getSearchWith } from '../../utils/searchHelper';
+import { SearchLink } from '../SearchLink/SearchLink';
+
+interface PeopleFiltersPerops {
```
volodymyr-soltys97, line 7: Typo in the interface name, it should be 'PeopleFiltersProps' instead of 'PeopleFiltersPerops'

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,127 @@
...
+          {people.map((person) => {
+            const selected = isSelected(person);
+
+            const findMother
+              = (): Person | undefined => {
+                return people
+                  .find((mother) => mother.name === person.motherName);
+              };
+
+            const findFather
```
volodymyr-soltys97, line 82: Avoid inline functions. It is better to declare 'findFather' outside the return statement and use the useCallback hook.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,127 @@
...
+              >
+                <td>
+                  <PersonLink person={person} />
+                </td>
+                <td>{person.sex}</td>
+                <td>{person.born}</td>
+                <td>{person.died}</td>
+
+                <td>
+                  {findMother() ? (
```
volodymyr-soltys97, line 102: Use optional chaining (findMother()?.slug) instead of checking if findMother() exists before accessing its properties.

volodymyr-soltys97 (changes requested):
Good job 🔥 
But let's make your code more perfect

src/components/Navigation/Navigation.tsx
AnastasiiaVorobets: if I import react lint throws an error

AnastasiiaVorobets: I don't understand why I need to do this.And I get errors with this syntax - 'export default const Navigation = () => (...)'.

volodymyr-soltys97 (approved):
Awesome 🚀

PR #325, opened 2023-06-28
add people-table-advanced

DidenkoDenis306:
[DEMO LINK](https://DidenkoDenis306.github.io/react_people-table-advanced/)


roman-mirzoian (changes requested):
The work is done well and the application itself works correctly, but there is a lot to work on in terms of optimization and readability of the code.
It is better to break large parts of the code into small pieces, then it will be easier for another person to understand this code, or even for you, when you open the code after a long time :)

PR #330, opened 2023-07-06
solution

ladrom:
[demo](https://ladrom.github.io/react_people-table-advanced/)


src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,24 @@
...
+};
+
+export const PersonLink: React.FC<Props> = ({ person }) => {
+  const location = useLocation();
+  const parentPath = useResolvedPath('../').pathname;
+
+  return (
+    <Link
+      to={{
+        pathname: parentPath + person.slug,
```
lerastarynets, line 16: The pathname is being concatenated with person.slug using '+'. It would be better to use a template literal here for better readability.

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,100 @@
...
+        ...person,
+        mother: peopleFromServer.find(
+          mother => mother.name === person.motherName,
+        ),
+        father: peopleFromServer.find(
+          father => father.name === person.fatherName,
+        ),
+      })));
+
+      if (!peopleFromServer) {
```
lerastarynets, line 37: This condition check should be placed before the 'setPeople' function is called, to prevent unnecessary operations when peopleFromServer is null or undefined.

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,100 @@
...
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    loadPeople();
+  }, []);
+
+  const filteredPeople = getFilteredPeople(people, searchParams);
```
lerastarynets, line 53: The 'searchParams' object is passed directly to the 'getFilteredPeople' function. This is a potential issue as the function is now tightly coupled with the search parameters structure. Consider passing explicit parameters instead.

lerastarynets, line 53: The filtering and sorting operations are performed on every render, which can be a performance issue if the people list is large. Consider using useMemo to optimize this.

src/App.tsx
```diff
@@ -1,20 +1,31 @@
...
+      <NavBar />
 
-      <div className="section">
+      <main className="section">
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
+            <Route path="*" element={<NotFoundPage />} />
```
lerastarynets, line 17: The `NotFoundPage` is set as the default route. It should only be displayed if no other route matches. Reorder the routes to fix this.

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,119 @@
...
+          </div>
+
+          <div className="level-right ml-4">
+            <SearchLink
+              data-cy="centuryALL"
+              className={
+                classNames(
+                  'button',
+                  'is-success',
+                  { 'is-outlined': centuriesParams.length !== 0 },
```
lerastarynets, line 94: pls [see falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)
```suggestion
                  { 'is-outlined': !!centuriesParams.length },
```

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/App.tsx
```diff
@@ -1,20 +1,31 @@
...
+            <Route path="/" element={<HomePage />} />
+            <Route path="home" element={<Navigate to="/" replace />} />
+            <Route path="*" element={<NotFoundPage />} />
+
+            <Route path="people">
+              <Route index element={<PeoplePage />} />
+              <Route path=":slug" element={<PeoplePage />} />
+            </Route>
+
+          </Routes>
```
polosanya, line 26:
```suggestion
            </Route>
          </Routes>
```
Don't leave empty lines between parent and child components

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { getPeople } from '../../api';
+import { PeopleTable } from '../PeopleTable';
+import { ErrorTypes } from '../../enum/ErrorTypes';
+import { PeopleFilters } from '../PeopleFilters/PeopleFilters';
+import { getFilteredPeople } from '../../utils/getFilteredPeople';
+import { getSortedPeople } from '../../utils/getSortedPeople';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [loading, setLoading] = useState(false);
```
polosanya, line 14:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```
image

src/utils/getSortedPeople.ts
```diff
@@ -0,0 +1,28 @@
...
+  params: Params,
+) {
+  let sortedPeople = [...people];
+  const { sort, order } = params;
+
+  if (sort === 'name' || sort === 'sex') {
+    sortedPeople.sort((a, b) => a[sort].localeCompare(b[sort]));
+  }
+
+  if (sort === 'born' || sort === 'died') {
```
polosanya, line 19: It is better to use switch...case here as you have limited values of sort

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments)

PR #339, opened 2023-07-13
solution

gabriella135:
[DEMO LINK](https://gabriella135.github.io/react_people-table-advanced/)

### DIFF
--- a/src/utils/getVisiblePeople.ts
+++ b/src/utils/getVisiblePeople.ts
@@ -1,12 +1,11 @@
 import { Person } from '../types';
-import { Sort } from '../types/Sort';
 import { sortPeople } from './sortPeople';
 
 interface Params {
-  sex: 'f' | 'm',
+  sex: string,
   query: string,
   centuries: string[],
-  sort: Sort,
+  sort: string,
   order: string,
 }
 
--- a/src/utils/sortPeople.ts
+++ b/src/utils/sortPeople.ts
@@ -1,26 +1,25 @@
 import { Person } from '../types';
-import { Sort } from '../types/Sort';
 
-export const sortPeople = (sortParam: Sort, people: Person[]) => {
+export const sortPeople = (sortParam: string, people: Person[]) => {
   const copy = [...people];
 
   switch (sortParam) {
-    case Sort.Name:
+    case 'name':
       return copy.sort((personA, personB) => (
         personA.name.localeCompare(personB.name)
       ));
 
-    case Sort.Sex:
+    case 'sex':
       return copy.sort((personA, personB) => (
         personA.sex.localeCompare(personB.sex)
       ));
 
-    case Sort.Born:
+    case 'born':
       return copy.sort((personA, personB) => (
         personA.born - personB.born
       ));
 
-    case Sort.Died:
+    case 'died':
       return copy.sort((personA, personB) => (
         personA.died - personB.died
       ));
--- a/src/types/Person.ts
+++ b/src/types/Person.ts
@@ -1,6 +1,6 @@
 export interface Person {
   name: string,
-  sex: 'm' | 'f',
+  sex: string,
   born: number,
   died: number,
   fatherName: string | null,
--- a/src/types/Sort.ts
+++ b/src/types/Sort.ts
@@ -1,6 +1,0 @@
-export enum Sort {
-  Name = 'name',
-  Sex = 'sex',
-  Born = 'born',
-  Died = 'died',
-}


src/utils/sortPeople.ts
```diff
@@ -0,0 +1,28 @@
+import { Person } from '../types';
+
+export const sortPeople = (sortParam: string, people: Person[]) => {
+  switch (sortParam) {
```
lerastarynets, line 4: The function 'sortPeople' is currently mutating the original 'people' array. This could lead to unexpected results in other parts of your application. To fix this, you should return a new sorted array instead of modifying the original one.

src/components/PeopleFilters.tsx
```diff
@@ -27,66 +64,51 @@ export const PeopleFilters = () => {
...
-            </a>
+            {centuries.map(century => (
+              <SearchLink
+                key={century}
+                data-cy="century"
+                className={classNames('button mr-1', {
+                  'is-info': centuryFilter.includes(century),
+                })}
+                params={{
+                  centuries: centuryFilter.includes(century)
```
lerastarynets, line 75: The use of ternary operators inside the parameters of `SearchLink` makes the code a bit hard to read. It would be more readable if you extract this logic into a separate function or variable.

src/components/PeopleTable.tsx
```diff
@@ -1,4 +1,45 @@
...
+    }
+
+    if (sort) {
+      return {
+        sort: sortParam,
+        order: 'desc',
+      };
+    }
+
+    return {
```
lerastarynets, line 37: When the 'sort' parameter does not exist, you should assign 'sortParam' to 'sort' and set 'order' to 'asc' as default.

src/components/PeopleTable.tsx
```diff
@@ -1,4 +1,45 @@
...
+  people,
+  findParent,
+}) => {
+  const [searchParams] = useSearchParams();
+  const sort = searchParams.get('sort');
+  const order = searchParams.get('order');
+  const { slug } = useParams();
+
+  const sortParams = (sortParam: string) => {
+    if (order && sort) {
```
lerastarynets, line 23: The 'sortParams' function logic is flawed. When 'order' and 'sort' parameters exist, they should not be set to null. Instead, 'order' should be toggled between 'asc' and 'desc'.

src/components/PeopleTable.tsx
```diff
@@ -1,4 +1,45 @@
...
+      return {
+        sort: null,
+        order: null,
+      };
+    }
+
+    if (sort) {
```
lerastarynets, line 30: When 'sort' parameter exists but 'order' does not, 'order' should be set to 'asc' as default, not 'desc'.

src/components/PeopleTable.tsx
```diff
@@ -56,627 +121,39 @@ export const PeopleTable = () => {
...
-            >
-              Maria Sturm
-            </a>
-          </td>
-          <td>
-            <a href="#/people/carolus-haverbeke-1832">Carolus Haverbeke</a>
-          </td>
-        </tr>
+        {
+          people.map(person => (
```
lerastarynets, line 125: You should add 'useCallback' for the function inside map to prevent unnecessary re-renders.

src/components/PeopleTable.tsx
```diff
@@ -9,44 +50,68 @@ export const PeopleTable = () => {
           <th>
```
lerastarynets, line 50: The sorting icon logic is incorrect. When 'order' is not defined, 'fa-sort' should be used. 'fa-sort-up' should be used when 'order' is 'asc', and 'fa-sort-down' when 'order' is 'desc'.

src/components/PeopleFilters.tsx
```diff
@@ -27,66 +64,51 @@ export const PeopleFilters = () => {
...
           </div>
 
           <div className="level-right ml-4">
-            <a
+            <SearchLink
               data-cy="centuryALL"
-              className="button is-success is-outlined"
-              href="#/people"
+              className={classNames('button is-success', {
+                'is-outlined': centuryFilter.length > 0,
```
lerastarynets, line 89: pls [see falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)
```suggestion
                'is-outlined': !!centuryFilter.length,
```

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,21 @@
+import { FC } from 'react';
+import { Link } from 'react-router-dom';
+import classNames from 'classnames';
+import { Person } from '../types';
+
+type Props = {
+  person : Person;
+};
+
+export const PersonLink: FC<Props> = ({ person }) => {
```
lerastarynets, line 10: Destucture `person`

src/utils/getVisiblePeople.ts
```diff
@@ -0,0 +1,56 @@
+import { Person } from '../types';
+import { sortPeople } from './sortPeople';
+
+interface Params {
+  sex: string,
```
lerastarynets, line 5: Consider using more specific types for 'sex' field in 'Params' interface. It seems like it should be a union type (e.g., 'male' | 'female') rather than a generic string.

src/utils/getVisiblePeople.ts
```diff
@@ -0,0 +1,56 @@
...
+  if (query) {
+    visiblePeople = visiblePeople.filter(person => {
+      const { name, fatherName, motherName } = person;
+      const lowerCaseQuery = (name + fatherName + motherName).toLowerCase();
+
+      return lowerCaseQuery.includes(query.toLowerCase());
+    });
+  }
+
+  if (centuries.length > 0) {
```
lerastarynets, line 39:
```suggestion
  if (centuries.length) {
```

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,108 @@
...
 
-              <p>There are no people matching the current search criteria</p>
+                  {!visiblePeople.length && (
+                    <p>
+                      There are no people matching the current search criteria
+                    </p>
+                  )}
 
-              <PeopleTable />
+                  {visiblePeople.length > 0 && (
```
lerastarynets, line 98:
```suggestion
                  {!!visiblePeople.length && (
```

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

witflash (changes requested):
Hi!
Almost good. But just one corner case.
If I click on someone's name, filters don't work anymore :(
And there is no change of highlight when I click any other name.
image

lerastarynets (changes requested):
Problem from prev comment is not fixed(
Everything breaks if I select a user. Pls ask for help in **fe_chat0** if you don't understand something

lerastarynets (changes requested):
Previous issue is fixed, however pls pay attention to task requirements. Now first click sorts a column descending, second click disables sorting while the required behaviour is
Screenshot at Jul 26 08-07-32

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,47 @@
+import classNames from 'classnames';
+import { useSearchParams } from 'react-router-dom';
+import { SearchLink } from './SearchLink';
+import { getSearchWith } from '../utils/searchHelper';
+
 export const PeopleFilters = () => {
+  const centuries = ['16', '17', '18', '19', '20'];
```
Anastasiia-Svintsova, line 7: You can move this array outside the component to not create it on each render

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,47 @@
...
+  const [searchParams, setSearchParams] = useSearchParams();
+  const sexFilter = searchParams.get('sex');
```
Anastasiia-Svintsova, line 9: Would be great to create an enum for search params to avoid errors connected with the typos in the future

src/components/PeopleFilters.tsx
Anastasiia-Svintsova: Not fixed 
image

src/components/PersonInfo.tsx
```diff
@@ -0,0 +1,46 @@
...
+      key={person.slug}
+      className={classNames({
+        'has-background-warning': slug === person.slug,
+      })}
+    >
+      <td>
+        <PersonLink person={person} />
+      </td>
+
+      <td>{person.sex}</td>
```
Anastasiia-Svintsova, line 27: Use destructuring for objects for a better readability

src/utils/getVisiblePeople.ts
Anastasiia-Svintsova: Not fixed. The sex can be 'f' or 'm'. Not just a random string
image

src/utils/sortPeople.ts
Anastasiia-Svintsova: Not fixed 
image

Anastasiia-Svintsova (changes requested):

maxim2310 (changes requested):
Make sure that you check out all previous mentors comments

witflash (approved):
Good job! 👍

PR #355, opened 2023-07-30
Develop

cebanoleksandr:
- DEMO LINK - https://cebanoleksandr.github.io/react_people-table-advanced/


src/components/PeoplePage.tsx
```diff
@@ -1,8 +1,42 @@
...
+  });
+
+  return preparedPeople;
+};
 
 export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [isEmpty, setIsEmpty] = useState(false);
```
anstsot, line 25: can be calculated based on `people` variable
```suggestion
  const isEmpty = !!people.length;
```

src/components/PeopleFilters.tsx
```diff
@@ -27,66 +67,48 @@ export const PeopleFilters = () => {
...
-            </a>
-
-            <a
-              data-cy="century"
-              className="button mr-1"
-              href="#/people?centuries=20"
-            >
-              20
-            </a>
+            {['16', '17', '18', '19', '20'].map(century => (
```
anstsot, line 70: better to create separate variable for filters

anstsot (changes requested):
Let’s add a few fixes to improve your code.
1. The text is shown on load
2. The sidebar with the filters should appear only when people are loaded.
image

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,91 @@
...
+                <p data-cy="noPeopleMessage">
+                  There are no people on the server
+                </p>
+              ) : (
+                <PeopleTable
+                  people={people}
+                  slug={slug}
+                  isLoading={isLoading}
+                />
+              )} */}
```
And678, line 88: Don't leave comments with old code.

src/components/PersonItem/PersonItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+      <td>
+        {person.father ? (
+          <Link
+            to={`../people/${person.father.slug}`}
+          >
+            {person.father.name}
+          </Link>
+        ) : (
+          person.fatherName || '-'
```
And678, line 62: Would be better to separate this code into `PeopleLink` component and reuse it

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,141 @@
...
+
+  const calculateSort = (value: string) => {
+    if (sort === value && order === 'desc') {
+      return '';
+    }
+
+    return value;
+  };
+
+  if (!isLoading && !preparePeople().length) {
```
And678, line 87: preparePeople() is a huge and heavy function, don't call it on every render just to check for length
```suggestion
  if (!isLoading && !preparePeople().length) {
```

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,141 @@
...
+        }
+      });
+    }
+
+    if (order === 'desc') {
+      newPeople = newPeople.reverse();
+    }
+
+    return newPeople;
+  };
```
And678, line 69: Memoize the result of this function instead of calling it in every render. (useMemo)

```suggestion
  const preparePeople = () => {
    let newPeople = [...people];

    if (query) {
      const normalizedQuery = query.trim().toLowerCase();

      newPeople = newPeople.filter(person => {
        return person.name.toLowerCase().includes(normalizedQuery)
          || (person.motherName || '').toLowerCase().includes(normalizedQuery)
          || (person.fatherName || '').toLowerCase().includes(normalizedQuery);
      });
    }

    if (sex) {
      newPeople = newPeople.filter(person => person.sex === sex);
    }

    if (centuries.length) {
      newPeople = newPeople.filter(person => {
        return centuries.includes(Math.ceil(person.born / 100).toString());
      });
    }

    if (sort) {
      newPeople = newPeople.sort((a, b) => {
        switch (sort) {
          case 'Name':
            return a.name.localeCompare(b.name);
          case 'Sex':
            return a.sex.localeCompare(b.sex);
          case 'Born':
            return a.born - b.born;
          case 'Died':
            return a.died - b.died;
          default:
            return 0;
        }
      });
    }

    if (order === 'desc') {
      newPeople = newPeople.reverse();
    }

    return newPeople;
  };
```

And678 (changes requested):
Good work, but can be even better :)

anstsot (approved):

PR #366, opened 2023-08-05
solution

MariiaMonakhova:
[DEMO LINK](https://MariiaMonakhova.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleLink.tsx
+++ b/src/components/PeopleLink.tsx
@@ -1,0 +1,87 @@
+import classNames from 'classnames';
+import React from 'react';
+import { Link, useParams, useSearchParams } from 'react-router-dom';
+import { Person } from '../types';
+
+type Props = {
+  person: Person,
+  onFindPerson: (name: string) => Person | undefined,
+};
+export const PeopleLink: React.FC<Props> = ({ person, onFindPerson }) => {
+  const { personSlug } = useParams();
+  const [searchParams] = useSearchParams();
+
+  return (
+    <tr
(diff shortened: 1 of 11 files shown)


src/components/PeopleLink.tsx
```diff
@@ -0,0 +1,87 @@
...
+            {person.motherName}
+          </Link>
+        )}
+
+        {person.motherName !== null
+        && onFindPerson(person.motherName) === undefined && (
+          `${person.motherName}`
+        )}
+
+        {!person.motherName && '-'}
```
artkas03, line 60: in tds for mother in father, you call `onFindPerson` function 3 times.
So to render one block this function has to find a person in the array 2 additional times even after you find it at the first check (line 40).
 
If we get to the math a bit... 
Imagine that we have 100 people array. In the worst case to render the parent block you are iterating the array 300 times instead of 100. Moreover, we have two such blocks. To render both blocks react iterates array 600 times instead of only 200. If we have 100 people in the array you can do 20 000 iterations (200 * 100) instead of 60 000 (600 * 100).

It would be great for the performance if you move results of both father and mother searching to variables and then use them.

artkas03 (approved):
opened your code to make clear about your recommendations and got a tip for you.
GJ though)

src/components/PeopleLink.tsx
MariiaMonakhova: Thank you for understanding and for the tip

anstsot (approved):

PR #373, opened 2023-08-08
feat: people table provided

OKhalanskyi:
[DEMO LINK](https://OKhalanskyi.github.io/react_people-table-advanced/)


src/App.tsx
```diff
@@ -1,20 +1,24 @@
...
-      <Navbar />
+    <Routes>
+      <Route path="/" element={<Layout />}>
+        <Route index element={<HomePage />} />
+
+        <Route path="people">
+          <Route path=":personSlug?" element={<PeoplePage />} />
+        </Route>
+
+        <Route path="home" element={<Navigate to="/" />} />
```
vladmakohon, line 18: The same as in the previous task
> add replace attribute not to save /home URL in the [browser history](https://reactrouter.com/en/main/start/tutorial#managing-the-history-stack) and avoid navigation loop when you press browser Go back button.

src/components/Navbar.tsx
```diff
@@ -8,15 +11,24 @@ export const Navbar = () => {
...
 
-          <a
+          <NavLink
             aria-current="page"
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
+            className={({ isActive }) => cn('navbar-item', {
+              'has-background-grey-lighter': isActive,
+            })}
+            to="/people"
```
vladmakohon, line 28: From the task description

> Keep search params when navigating within the People page (when selecting a person or clicking the People link).

Tip: Here, you can get the current `searchParams` and pass the object with the corresponding `pathname` and `search` properties.

vladmakohon (changes requested):
It looks like you forgot to commit some files. However, I can already see that some things can be fixed.

PR #379, opened 2023-08-09
solution

dianadiv:
[DEMO LINK](https://dianadiv.github.io/react_people-table-advanced/)


anstsot (changes requested):
Let’s add a few fixes to improve your code. Fix one bug:
- Text should be added once there are no filtered people
image

- the close icon doesn't work and there is odd behavior of input sometimes
image

anstsot (approved):

PR #386, opened 2023-08-09
Solution

outofs:
https://outofs.github.io/react_people-table-advanced/


src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,67 @@
...
+
+import classNames from 'classnames';
+import { useSearchParams } from 'react-router-dom';
+import { getSearchWith } from '../utils/searchHelper';
+import { SearchLink } from './SearchLink';
+
+const allCenturies = ['16', '17', '18', '19', '20'];
+
+enum Sex {
+  All = 'all',
```
perfectAcedia, line 11: This type is unused, so I guess you can remove "All".

perfectAcedia (approved):
Looks good, but I found a few issues. 
Firstly, switch pages between Home and People, you will see short table loading, try to remove it.
Then, I found that "arrow sorting" saves sort params in the URL but does not remove them once sorting is disabled. Also, after the first click on the sort button, both(sort and order) fields appear. It shouldn't work like this I guess.
First click -> only sort field
Second click -> sort + order field
Third click -> both fields should be removed from the URL bar.
image

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  useEffect(() => {
+    setLoading(true);
+    getPeople()
+      .then(setPeople)
+      .catch(() => setIsError(true))
+      .finally(() => setLoading(false));
+  }, []);
+
+  const preparedPeople = preparePeople(
```
anstsot, line 33: `useMemo` can be used

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,67 @@
+import React from 'react';
+
+import classNames from 'classnames';
+import { useSearchParams } from 'react-router-dom';
+import { getSearchWith } from '../utils/searchHelper';
+import { SearchLink } from './SearchLink';
+
+const allCenturies = ['16', '17', '18', '19', '20'];
+
+enum Sex {
```
anstsot, line 10: better to move in `typings` folder

src/components/PeopleRow.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const foundPerson = table.find(person => person.name === name);
+
+  return foundPerson
+    ? <PersonLink person={foundPerson} />
+    : name;
+};
+
+export const PeopleRow: React.FC<Props> = ({ person, slug, people }) => {
+  const linkedMother = useMemo(
+    () => findLinkedPerson(person.motherName || '', people), [],
```
anstsot, line 23: add missing deps
```suggestion
    () => findLinkedPerson(person.motherName || '', people), [people],
```

anstsot (changes requested):
Let’s add a few fixes to improve your code.
- Test should be visible onle
image

anstsot (approved):

PR #391, opened 2023-08-09
solution

olegpaliy:
https://olegpaliy.github.io/react_people-table-advanced/


src/components/PeopleFilters.tsx
```diff
@@ -1,92 +1,118 @@
...
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+  const sex = searchParams.get('sex') || '';
+
+  const centuryOptions = ['16', '17', '18', '19', '20'];
+
+  const sexOptions = [
+    { title: 'Male', value: 'm' },
+    { title: 'Female', value: 'f' },
+  ];
```
anstsot, line 18: better to move outside the component

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,99 @@
...
+  const order = searchParams.get('order') || '';
+
+  const sortOptions = [
+    { title: 'Name', value: 'name' },
+    { title: 'Sex', value: 'sex' },
+    { title: 'Born', value: 'born' },
+    { title: 'Died', value: 'died' },
+  ];
+
+  const findPerson = (name: string | null) => {
```
anstsot, line 24: use `useCallback`

anstsot (changes requested):
But pay attention to my comments to avoid the same mistakes in the future
- show text only
image

anstsot (approved):
But pay attention to my comments to avoid the same mistakes in the future
- show text only
image

src/components/PeopleTable.tsx
olegpaliy: fixed)

PR #414, opened 2023-08-11
add task solution

NadiaKovalchuk:
[DEMO LINK] https://NadiaKovalchuk.github.io/react_people-table-advanced/


src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const sex = searchParams.get('sex') || '';
+  const centuries = searchParams.getAll('centuries') || '';
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getPeople()
+      .then((peopleFromServer) => {
+        setPeople(peopleFromServer);
+      })
```
artkas03, line 28: here you could pass just set function to then
`getPeople().then(setPeople)`

src/utils/getPreparedPeople.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  if (query) {
+    preparedPeople = preparedPeople.filter((person) => {
+      return (
+        person.name.toLowerCase().includes(query.toLowerCase().trim())
+        || person.motherName?.toLowerCase().includes(query.toLowerCase().trim())
+        || person.fatherName?.toLowerCase().includes(query.toLowerCase().trim())
+      );
+    });
+  }
```
artkas03, line 18: it is much better to make variable for normalized query to avoid transforming and triming query for every person for 3 times
line 11: `const normalazidQuery = query.toLowerCase().trim()`
and then use it

artkas03 (approved):
GJ)
left a few tips for you)
also this buttons are not working
image

src/App.tsx
```diff
@@ -1,20 +1,15 @@
-import { PeoplePage } from './components/PeoplePage';
-import { Navbar } from './components/Navbar';
-
+import { Outlet } from 'react-router-dom';
 import './App.scss';
+import NavBar from './components/NavBar';
```
And678, line 3: That file is called `Navbar.tsx`
```suggestion
import NavBar from './components/Navbar';
```

And678: Same for other imports

And678 (changes requested):
Sorting doesn't work as @artkas03 pointed out. Also you need to fix test/lint, partially this is because of broken imports.

PR #416, opened 2023-08-13
task done

pos1bl:
https://pos1bl.github.io/react_people-table-advanced/

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -1,0 +1,26 @@
+import {
+  Navigate,
+  Route,
+  HashRouter as Router,
+  Routes,
+} from 'react-router-dom';
+
+import { App } from './App';
+import { HomePage } from './components/HomePage';
+import { PeoplePage } from './components/PeoplePage';
+import { InvalidPage } from './components/InvalidPage';
+
+export const Root = () => (
+  <Router>
+    <Routes>
+      <Route path="/" element={<App />}>
+        <Route index element={<HomePage />} />
+        <Route path="home" element={<Navigate to="/" />} />
+        <Route path="people" element={<PeoplePage />}>
+          <Route path=":personSlug" />
+        </Route>
+        <Route path="*" element={<InvalidPage />} />
+      </Route>
+    </Routes>
+  </Router>
+);
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,21 +1,68 @@
-export const PeopleFilters = () => {
+import classNames from 'classnames';
+import { SearchLink } from './SearchLink';
+import { CenturiesOptions, SexOptions } from '../utils/data';
+
+type Props = {
+  sex: SexOptions;
+  centuries: CenturiesOptions[];
+  query: string;
+  handleQueryChange: (event: React.ChangeEvent<HTMLInputElement>) => void,
+};
+
+export const PeopleFilters: React.FC<Props> = ({
+  sex,
+  centuries,
+  query,
+  handleQueryChange,
+}) => {
+  const centuryParams = (century: CenturiesOptions) => {
+    return centuries.includes(century)
+      ? centuries.filter(num => century !== num)
+      : [...centuries, century];
+  };
+
+  const resetParams = {
+    centuries: null,
+    sex: null,
+    query: null,
+  };
+
   return (
     <nav className="panel">
       <p className="panel-heading">Filters</p>
 
       <p className="panel-tabs" data-cy="SexFilter">
-        <a className="is-active" href="#/people">All</a>
-        <a className="" href="#/people?sex=m">Male</a>
-        <a className="" href="#/people?sex=f">Female</a>
+        <SearchLink
+          className={classNames({ 'is-active': !sex })}
+          params={{ sex: null }}
+        >
+          All
+        </SearchLink>
+
+        <SearchLink
+          className={classNames({ 'is-active': sex === SexOptions.MALE })}
+          params={{ sex: SexOptions.MALE }}
+        >
+          Male
+        </SearchLink>
+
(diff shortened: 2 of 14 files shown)


src/Root.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { InvalidPage } from './components/InvalidPage';
+
+export const Root = () => (
+  <Router>
+    <Routes>
+      <Route path="/" element={<App />}>
+        <Route index element={<HomePage />} />
+        <Route path="home" element={<Navigate to="/" />} />
+        <Route path="people" element={<PeoplePage />}>
+          <Route path=":personSlug" />
```
anastasiiavorobiova, line 20: The nested route with ':personSlug' seems to be missing an element. You should specify a component that will be rendered when navigating to '/people/:personSlug'.

src/components/PeopleFilters.tsx
```diff
@@ -1,21 +1,68 @@
...
+  query,
+  handleQueryChange,
+}) => {
+  const centuryParams = (century: CenturiesOptions) => {
+    return centuries.includes(century)
+      ? centuries.filter(num => century !== num)
+      : [...centuries, century];
+  };
+
+  const resetParams = {
```
anastasiiavorobiova, line 24: The resetParams object should not contain null values. Instead, you should use appropriate initial values for centuries, sex, and query. For example, an empty array for centuries, an empty string for sex and query.

src/components/PeopleFilters.tsx
```diff
@@ -1,21 +1,68 @@
...
+          Female
+        </SearchLink>
       </p>
 
       <div className="panel-block">
         <p className="control has-icons-left">
           <input
             data-cy="NameFilter"
-            type="search"
+            type="text"
```
anastasiiavorobiova, line 61: The input type should be 'search', not 'text'. The 'search' type is specially designed to make search boxes more user-friendly.

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,151 @@
...
-              </p>
+              {tableValidate && !people.length && (
+                <p data-cy="noPeopleMessage">
+                  There are no people on the server
+                </p>
+              )}
 
-              <p>There are no people matching the current search criteria</p>
+              {tableValidate && !currentPeople.length && !!people.length && (
+                <p data-cy="noPeopleMessage">
```
anastasiiavorobiova, line 137: The message 'There are no people on the server' is misleading. The message should be 'There are no people matching the current search criteria'.

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,151 @@
...
+  useEffect(() => {
+    setIsLoading(true);
+
+    getPeople()
+      .then(setPeople)
+      .catch(() => setIsError(true))
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  useMemo(() => {
```
anastasiiavorobiova, line 40: Using useMemo without returning a value is an anti-pattern. useMemo is meant to memoize expensive computations. Consider splitting this logic into a function and call this function when needed.

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,151 @@
...
+
+        return 0;
+      });
+    }
+
+    setCurrentPeople(newList);
+  },
+  [people, searchParams]);
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
anastasiiavorobiova, line 92: You are performing trim operation on a value that you are not using in case there is a trim value. Consider using trimmed value instead of the original one.

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,73 @@
...
-            </span>
-          </th>
-
-          <th>
-            <span className="is-flex is-flex-wrap-nowrap">
-              Died
-              <a href="#/people?sort=died">
-                <span className="icon">
-                  <i className="fas fa-sort" />
+          {Object.values(SortOptions).map(option => {
```
anastasiiavorobiova, line 38: To make your code cleaner, you can use the implicit return of the arrow function. Like this: {Object.values(SortOptions).map(option => (<th key={option}.../>))}

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,73 @@
...
+            return (
+              <th key={option}>
+                <span className="is-flex is-flex-wrap-nowrap">
+                  {option.charAt(0).toUpperCase() + option.slice(1)}
```
anastasiiavorobiova, line 42: To follow best practices, you should create a function to capitalize the first letter of a string and use it here.

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,73 @@
...
+
+type Props = {
+  people: Person[];
+  sort: SortOptions;
+  order: OrderOption;
+};
+
+export const PeopleTable: React.FC<Props> = ({
+  people,
+  sort,
```
anastasiiavorobiova, line 16: The 'sort' and 'order' from props should have default values in case they are not provided.

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/components/PeopleFilters.tsx
pos1bl: image
If i change null to initial values after Reset i will get 
image
I think its not correct )

src/Root.tsx
pos1bl: There is no point in adding the same element, because it does not change with the parent, and I use a nested route, because since version 6.* the ability to use the optional path parameterrh has been removed

volodymyr-soltys97 (approved):
Well done 🔥

PR #420, opened 2023-08-14
Develop

ArtemAndrushchenko:
[DEMO LINK](https://ArtemAndrushchenko.github.io/react_people-table-advanced/)


anstsot (changes requested):
One condition from the task is not met
- add sorting possibility
image

PR #434, opened 2023-08-24
first commit

Kocha1991:
[DEMO LINK](https://Kocha1991.github.io/react_people-table-advanced/)


src/components/PeopleFilters.tsx
```diff
@@ -16,77 +77,61 @@ export const PeopleFilters = () => {
...
 
-          <span className="icon is-left">
-            <i className="fas fa-search" aria-hidden="true" />
+          <span
+            className="icon is-left"
+          >
+            <i
+              className="fas fa-search"
+              aria-hidden="true"
+              onClick={() => clearCentury}
```
IvanFesenko, line 90:
```suggestion
              onClick={clearCentury}
```

IvanFesenko (changes requested):
Sorting works wrong it should work in both cases + default state

2023-08-24_15h18_43

VitaliyBondarenko1982 (changes requested):
Almost done.
Please check such behavior (on sort appears loader which make some jumpings) and compare it with [working example](https://mate-academy.github.io/react_people-table-advanced/#/people?sort=born)
chrome-capture-2023-8-21 (3)

sTorba24 (approved):

PR #440, opened 2023-08-31
Solution

Mar4enkofff:
https://Mar4enkofff.github.io/react_people-table-advanced/


src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <div className="columns is-desktop is-flex-direction-row-reverse">
+          {loading && <Loader />}
+
+          {error && (
+            <p data-cy="peopleLoadingError">
+              Something went wrong
+            </p>
+          )}
+
+          {!people && (
```
TarasHoliuk, line 74: the test fails because if there are no people you do this: `![]`. But an empty array is a trusty value (i.e. !![] === true)
image

TarasHoliuk (changes requested):
Very good job :+1: 
Added one comment to fix tests
P.S. When you fix a test - check if other tests pass

PR #443, opened 2023-09-02
react_people-table-advanced

vintwp:
https://vintwp.github.io/react_people-table-advanced/#/


volodymyr-soltys97 (changes requested):
Good job 👍 
No need to show this block when people is loading on the page, need to show after loading
image

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2023-09-03 at 17 11 38
Clicking sort causes resetting all checked filters

nazarmatsevych (approved):
👍

PR #444, opened 2023-09-02
Solution

s3rg4sh:
https://s3rg4sh.github.io/react_people-table-advanced/


src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,106 @@
+import { useSearchParams } from 'react-router-dom';
+import cn from 'classnames';
+import { SearchLink } from '../SearchLink';
+import { getSearchWith } from '../../utils/searchHelper';
+
+export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const sex = searchParams.get('sex');
+  const centuriesParams = searchParams.getAll('centuries');
```
etojeDenys, line 9: let's move this query param names to some enum, it will help us to manipulate them easily

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const lowerQuery = query.toLocaleLowerCase();
+
+    preparedPeople = preparedPeople.filter(person => {
+      return [
+        person.name,
+        person.motherName || '',
+        person.fatherName || '',
+      ]
+        .join('\n')
+        .toLocaleLowerCase()
```
etojeDenys, line 47: let's move this logic to some function, it will look better and be more readable

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const getPersonParent = (name: string | null): JSX.Element => {
+    return (name ? (
+      <>
+        {getPersonByName(name)
+          ? (
+            <Personlink
+              person={getPersonByName(name) as Person}
+            />
+          )
```
etojeDenys, line 26: let's save the function result to avoid redundant function call

etojeDenys (changes requested):

src/components/PeopleFilters/PeopleFilters.tsx
etojeDenys: still not fixed
image

feel free to ask in the chat if you have some questions

etojeDenys (changes requested):

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+          >
+            <td>
+              <Personlink person={person} />
+            </td>
+
+            <td>{person.sex}</td>
+            <td>{person.born}</td>
+            <td>{person.died}</td>
+            <td>
+              {getPersonParent(person.motherName)}
```
VitaliyBodnar, line 98: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,150 @@
...
+          <div className="column is-7-tablet is-narrow-desktop">
+            <PeopleFilters />
+          </div>
+
+          <div className="column">
+            <div className="box table-container">
+              {isLoading ? (
+                <Loader />
+              ) : (
+                loadPeopleTable()
```
VitaliyBodnar, line 142: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable
please check other places

VitaliyBodnar (approved):

PR #454, opened 2023-09-09
solution

vlkzmn:
[DEMO LINK](https://vlkzmn.github.io/react_people-table-advanced/)


src/components/TableHead.tsx
```diff
@@ -0,0 +1,57 @@
...
+      sortParams.sort = null;
+      sortParams.order = null;
+    }
+
+    return sortParams;
+  }
+
+  return (
+    <tr>
+      {['name', 'sex', 'born', 'died'].map(field => {
```
lerastarynets, line 29: Avoid magic strings. It would be better to define the array ['name', 'sex', 'born', 'died'] as a constant above the component.

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

Anastasiia-Svintsova (approved):
👍

PR #461, opened 2023-09-20
add task solution

UlianaChorna:
[DEMO LINK](https://UlianaChorna.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,12 +1,75 @@
+import React from 'react';
+import classNames from 'classnames';
+import { useSearchParams } from 'react-router-dom';
+import { getSearchWith } from '../utils/searchHelper';
+import { SearchLink } from './SearchLink';
+
+const availableCenturies = ['16', '17', '18', '19', '20'];
+
 export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const query = searchParams.get('query') || '';
+  const sex = searchParams.get('sex') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+
+  if (availableCenturies.length === centuries.length) {
+    setSearchParams(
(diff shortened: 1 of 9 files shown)


denys-danyliuk (changes requested):
this buttons work only after second click
Знімок екрана 2023-09-20 о 14 37 46

all female names should have red color, and male - blue
Знімок екрана 2023-09-20 о 14 36 47

you should remove this text when there are people in the list, also when you go from home page to people page this text also appears when loader is active, check how it should work and change it
Знімок екрана 2023-09-20 о 14 36 24

src/components/PeopleFilters.tsx
```diff
@@ -16,6 +79,8 @@ export const PeopleFilters = () => {
             type="search"
             className="input"
             placeholder="Search"
+            value={query}
+            onChange={(e) => handleQueryChange(e)}
```
TarasHoliuk, line 83:
```suggestion
            onChange={handleQueryChange}
```
you should pass a function that accepts the event as an argument - like your `handleQueryChange` function, the additional wrapper is redundant - FYI

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,123 @@
...
+  const visiblePerson = useMemo(() => {
+    let temp = people.map(person => ({
+      ...person,
+      mother: people.find(mother => mother.name === person.motherName) || null,
+      father: people.find(father => father.name === person.fatherName) || null,
+    }));
+
+    if (query) {
+      temp = temp.filter(person => (
+        normalized(person.name).includes(normalized(query))
```
TarasHoliuk, line 47: you might save the `normalized(query)` into a variable to prevent doing it every time - FYI

TarasHoliuk (approved):
Well done :+1: 
Added 2 comments just FYI

PR #466, opened 2023-09-27
task solution

oleksandrshtonda:
[DEMO LINK](https://oleksandrshtonda.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,12 +1,47 @@
+import { useSearchParams } from 'react-router-dom';
+import cn from 'classnames';
+import { SearchLink } from './SearchLink';
+import { getSearchWith } from '../utils/searchHelper';
+import { SearchParam } from '../types/SearchParam';
+
+const CENTURIES = ['16', '17', '18', '19', '20'];
+
 export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const sex = searchParams.get(SearchParam.sex);
+  const centuriesParams = searchParams.getAll(SearchParam.centuries);
+
+  const queryChangeHandler = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchParams(
+      getSearchWith(searchParams, {
+        query: e.target.value || null,
+      }),
+    );
+  };
+
   return (
     <nav className="panel">
       <p className="panel-heading">Filters</p>
 
       <p className="panel-tabs" data-cy="SexFilter">
-        <a className="is-active" href="#/people">All</a>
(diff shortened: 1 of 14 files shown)


volodymyr-soltys97 (changes requested):
This filter should only be shown after people have downloaded it
image

src/components/PeopleFilters.tsx
```diff
@@ -27,66 +63,46 @@ export const PeopleFilters = () => {
...
+                className={cn(
+                  'button mr-1',
+                  {
+                    'is-info': centuriesParams.includes(century),
+                  },
+                )}
+                params={{
+                  centuries: centuriesParams.includes(century)
+                    ? centuriesParams.filter(cent => cent !== century)
+                    : [...centuriesParams, century],
```
lerastarynets, line 79: pls move long expressions to separate consts

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,146 @@
...
+  sortField: string | null,
+  order: string | null,
+): Person[] {
+  let preparedPeople = [...people];
+
+  if (sex) {
+    preparedPeople = preparedPeople.filter(person => person.sex === sex);
+  }
+
+  if (centuries.length > 0) {
```
lerastarynets, line 24:
```suggestion
  if (!!centuries.length) {
```

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,146 @@
...
+      <>
+        {isError
+          ? (
+            <p data-cy="peopleLoadingError" className="has-text-danger">
+              Something went wrong
+            </p>
+          )
+          : (
+            <>
+              {people.length > 0
```
lerastarynets, line 105:
```suggestion
              {!!people.length
```

lerastarynets (changes requested):
Almost there) Just several more things needs to be fixed

src/components/PeoplePage.tsx
oleksandrshtonda: I can't type !! before centuries.length because ESlint show error "Redundant double negation.eslint". So in the next commit I'll type just  centuries.length

src/components/PeoplePage.tsx
oleksandrshtonda: I can't type !! before people.length because ESlint show error "Redundant double negation.eslint". So in the next commit I'll type just  people.length

anastasiiavorobiova (changes requested):
Great work! Almost done: the reset button also should clear the search input field

IvanFesenko (approved):
well done

PR #468, opened 2023-09-28
realized: all filters and sort params saving as url params, filtering…

DenisChakhalian:
[DEMO LINK](https://denischakhalian.github.io/react_people-table-advanced/)


Wesses (approved):
LGTM!

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,83 @@
+import { useContext, useEffect, useState } from 'react';
+import { getPeople } from '../../api';
+import { Loader } from '../Loader';
+import { PeopleList } from '../PeopleList/PeopleList';
+import { PeopleFilters } from '../PeopleFilters/PeopleFilters';
```
alexander-ignatow, line 5: Such imports look bit ugly

Why not rename `PeopleFilters/PeopleFilters.tsx` to just `PeopleFilters/index.tsx`

Than you will get :

```suggestion
import { PeopleFilters } from '../PeopleFilters';
```

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,83 @@
...
+      .then(currentPeople => {
+        setPeople(getPreparedPeople(currentPeople));
+      })
+      .catch(() => {
+        setIsError(true);
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, []);
```
alexander-ignatow, line 28: From one side this looks correct
But - open the People tab, then Home, then People again

When opening the tab for a second time you have a "flash" of the old people table, then it disappears and loading comes, then new content appears.

It would be perfect to work around this unpleasant UX
For instance - you can check if you have already loaded some people then show the old content, but in some "disabled" mode - to prevent selecting a person, while you have not synchronized with the server. Or you may clear the people when leaving the People tab - but neither solution is simple.

No need to fix this here - just wanted to point you to some real-world problems that may potentially appear with complex interfaces.

src/variables.ts
alexander-ignatow, line 1: These are constants, not variables :)

alexander-ignatow (changes requested):
Very good! 👍 

I would point out just one issue with how you arrange your components in the folder

For now - all component files/folders are flat-listed in the components folder.
Can you please re-arrange them a bit? For instance: you may create a folder for the page-related components, or create folders and group components based on the page they belong to, etc. The point here is to make your solution easier to maintain.

Another issue is that you have files like `HomePage/HomePage.tsx` - which leads to a bit of ugly imports. If you rename the "entry points" of your components to `index` - you will be able to import them in a shorter way.

src/utils/peopleHelper.ts
```diff
@@ -0,0 +1,11 @@
+import { Person } from '../types';
+
+export function getPreparedPeople(people: Person[]) {
+  return people.map(person => {
+    return {
+      ...person,
+      mother: people.find(mother => mother.name === person.motherName),
+      father: people.find(father => father.name === person.fatherName),
```
roman-mirzoian, line 8:
```suggestion
      mother: people.find(({ name }) => name === person.motherName),
      father: people.find(({ name }) => name === person.fatherName),
```

roman-mirzoian (approved):
Well done 👍

PR #493, opened 2023-10-03
add initial solution

desevoker:
[DEMO LINK](https://desevoker.github.io/react_people-table-advanced/)


serkrops (approved):
Awesome! Looks very good!

Moroz-Dmytro (changes requested):
image

src/api.ts
```diff
@@ -1,8 +1,6 @@
+import { API_URL } from './utils/constants';
 import { Person } from './types/Person';
 
-// eslint-disable-next-line max-len
-const API_URL = 'https://mate-academy.github.io/react_people-table/api/people.json';
```
BudnikOleksii, line 4: You can keep it here, as this constant related to API only

src/components/PeopleFilters.tsx
```diff
@@ -1,21 +1,67 @@
...
+    const params = new URLSearchParams(searchParams);
+    const inputValue = event.target.value;
+
+    if (inputValue) {
+      params.set(SearchParam.Query, inputValue);
+    } else {
+      params.delete(SearchParam.Query);
+    }
+
+    setSearchParams(params);
```
BudnikOleksii, line 26: You can get and modify previous search params with setSearchParams.
Something like this:
```javascript
setSearchParams((prevParams) => {
  if (inputValue) {
    prevParams.set(SearchParam.Query, inputValue);
  } else {
    prevParams.delete(SearchParam.Query);
  }

  return prevParams;
});
```

src/components/SortLink.tsx
```diff
@@ -0,0 +1,44 @@
...
+  if (sort !== field) {
+    params.sort = field;
+
+    iconClass = 'fa-sort';
+  } else if (sort === field && !order) {
+    params.sort = field;
+    params.order = SortOrder.Desc;
+
+    iconClass = 'fa-sort-up';
+  }
```
BudnikOleksii, line 33: You can create a function that returns `iconClass` and move it outside the component

BudnikOleksii (approved):
Nicely done 🔥

PR #501, opened 2023-10-04
react_people-table-advanced

vatatan:
[DEMO LINK](https://vatatan.github.io/react_people-table-advanced/)


voievodik (changes requested):
Good job! But there are some bags in your solution. [Check out the working example.](https://github.com/mate-academy/react_people-table-advanced)

You need to set the correct validation to change the style for the sex filter:
[https://gyazo.com/8841cd125a85027035b144fb3b8ca58d](https://gyazo.com/8841cd125a85027035b144fb3b8ca58d)

Also, your sorting doesn't work properly, you need to sort by field first, then reverse the sort and on the third click to clear it. Your icons need to be changed according to your sorting field.
[https://gyazo.com/61b1555ea79704117572255fdcb9b031](https://gyazo.com/61b1555ea79704117572255fdcb9b031)

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,84 @@
...
             <div className="box table-container">
-              <Loader />
-
-              <p data-cy="peopleLoadingError">Something went wrong</p>
+              {loading && !hasError && !people.length && <Loader />}
 
-              <p data-cy="noPeopleMessage">
-                There are no people on the server
-              </p>
+              {!loading && hasError && !people.length && (
```
voievodik, line 68: It would be better to separate this check to the variable. And also for other ones.

src/components/HeadOfTable.tsx
```diff
@@ -0,0 +1,38 @@
+import classNames from 'classnames';
+import React from 'react';
+import { NavLink } from 'react-router-dom';
+
+export const HeadOfTable: React.FC = () => {
+  const handleNavLinkClass = ({ isActive }: { isActive: boolean }) => {
+    return classNames('navbar-item', {
+      'has-background-grey-lighter': isActive,
+    });
+  };
```
roman-mirzoian, line 10: It is better to define this function outside the component so that it is not created every time a new render is performed.

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,61 @@
...
+import { getSearchWith } from '../utils/searchHelper';
+import { SearchLink } from './SearchLink';
+import { Sex } from '../types/Sex';
+
+type Props = {
+  people: Person[]
+};
+
+export const PeopleFilters: React.FC<Props> = () => {
+  const CENTURIES = ['16', '17', '18', '19', '20'];
```
roman-mirzoian, line 16: It is better to define this array outside the component, so that it is not created every time during a new rendering.

src/components/PeoplePage.tsx
roman-mirzoian: Good point

src/utils/filterPeople.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export const filterPeople = (
+  people: Person[],
+  sex: string | null,
+  query: string | null,
+  centuries: string[],
+  sortField: string | null,
+  order: string | null,
+) => {
+  let tempPeople = [...people];
```
roman-mirzoian, line 11: The `filter` method returns a copy of the array, so you don't need to do it manually.

src/utils/filterPeople.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  if (sex) {
+    tempPeople = tempPeople.filter(person => person.sex === sex);
+  }
+
+  if (centuries.length) {
+    const getCentury = (person: Person) => Math.ceil(person.born / 100);
```
roman-mirzoian, line 18: Avoid using "magic values" (100 is a magic value)

src/utils/filterPeople.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    tempPeople = tempPeople.filter(
+      person => centuries.includes(
+        getCentury(person).toString(),
+      ),
+    );
```
roman-mirzoian, line 24: Do not shorten the code entry to difficult to read, it is better to break it into logical parts.
You can also immediately return a `string` value from the `getCentury` function.
```js
const getCentury = (person: Person) => Math.ceil(person.born / 100).toString();
```
```suggestion
    tempPeople = tempPeople.filter(person => {
      const personCentury = getCentury(person);
      return centuries.includes(personCentury);
     });
```

src/utils/filterPeople.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  if (query) {
+    const lowerQuery = query.toLocaleLowerCase();
+
+    tempPeople = tempPeople.filter((person) => {
+      const searchableStrings = [
+        person.name,
+        person.motherName || '',
```
roman-mirzoian, line 33:
```js
const NOT_SET_VALUE = '-';
...
person.motherName || NOT_SET_VALUE,
```

roman-mirzoian (changes requested):
To the problems with styles already described by Taras, I can add the problem of changing the width of the table during filtering:
https://gyazo.com/315e4a428a0339f7a0444223718eeb73

PR #513, opened 2023-10-04
Develop

Yevheniia-Sid:
[DEMO LINK](https://Yevheniia-Sid.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,12 +1,46 @@
+import classNames from 'classnames';
+import { Link, useSearchParams } from 'react-router-dom';
+import { getSearchWith } from '../utils/searchHelper';
+import { CENTURIES_LINK } from '../utils/variables';
+import { Gender } from '../types/Gender';
+
 export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const query = searchParams.get('query') || '';
+  const sex = searchParams.get('sex') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+
+  function handleQueryChange(event: React.ChangeEvent<HTMLInputElement>) {
+    const params = { query: event.target.value || null };
+    const search = getSearchWith(searchParams, params);
+
+    setSearchParams(search);
+  }
+
   return (
     <nav className="panel">
       <p className="panel-heading">Filters</p>
 
       <p className="panel-tabs" data-cy="SexFilter">
-        <a className="is-active" href="#/people">All</a>
-        <a className="" href="#/people?sex=m">Male</a>
-        <a className="" href="#/people?sex=f">Female</a>
+        <Link
+          className={classNames({ 'is-active': !sex })}
+          to={{ search: getSearchWith(searchParams, { sex: null }) }}
+        >
+          All
+        </Link>
+        <Link
+          className={classNames({ 'is-active': sex === Gender.Male })}
+          to={{ search: getSearchWith(searchParams, { sex: Gender.Male }) }}
+        >
+          Male
+        </Link>
+        <Link
+          className={classNames({ 'is-active': sex === Gender.Female })}
+          to={{ search: getSearchWith(searchParams, { sex: Gender.Female }) }}
+        >
+          Female
+        </Link>
       </p>
 
       <div className="panel-block">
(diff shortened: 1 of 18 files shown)


src/components/PeopleFilters.tsx
```diff
@@ -27,66 +65,51 @@ export const PeopleFilters = () => {
...
           <div className="level-right ml-4">
-            <a
+            <SearchLink
               data-cy="centuryALL"
-              className="button is-success is-outlined"
-              href="#/people"
+              className={classNames(
+                'button',
+                'is-success',
+                { 'is-outlined': centuries.length !== 0 },
```
roman-mirzoian, line 97:
```suggestion
                { 'is-outlined': !!centuries.length },
```

src/utils/FilterSortPeople.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  if (sexFilter) {
+    filteredPeople = filteredPeople
+      .filter(({ sex }) => sex === sexFilter);
+  }
+
+  if (centuriesFilter.length) {
+    filteredPeople = filteredPeople
+      .filter(({ born }) => centuriesFilter
+        .includes((Math.ceil(born / 100)).toString()));
```
roman-mirzoian, line 36: Avoid using "magic values" (100)

src/utils/FilterSortPeople.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  if (sortField) {
+    switch (sortField) {
+      case SortFields.Name:
+        filteredPeople
+          .sort((personA, personB) => personA.name.localeCompare(personB.name));
```
roman-mirzoian, line 43:
```suggestion
          .sort((a, b) => a.name.localeCompare(b.name));
```

src/utils/FilterSortPeople.ts
```diff
@@ -0,0 +1,68 @@
...
+  people: Person[],
+  queryParams: URLSearchParams,
+): Person[] => {
+  const query = queryParams.get(QueryParamsType.Query) || '';
+  const sexFilter = queryParams.get(QueryParamsType.Sex);
+  const centuriesFilter = queryParams.getAll(QueryParamsType.Centuries);
+  const sortField = queryParams.get(QueryParamsType.Sort);
+  const order = queryParams.get(QueryParamsType.Order);
+
+  let filteredPeople = [...people];
```
roman-mirzoian, line 15: The `filter` method returns a copy of the array, so you don't need to do it manually.

src/utils/getPreparedPeople.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const getPreparedPeople = (peopleFromServer: Person[]) => {
+  return peopleFromServer
+    .map(person => {
+      const tempPerson = person;
+
+      tempPerson.motherName = person.motherName || NOT_SET_VALUE;
+      tempPerson.fatherName = person.fatherName || NOT_SET_VALUE;
+      tempPerson.mother = peopleFromServer
+        .find(mother => mother.name === person.motherName);
```
roman-mirzoian, line 12:
```suggestion
      tempPerson.mother = peopleFromServer
        .find(({ name }) => name === person.motherName);
```

roman-mirzoian (changes requested):

Moroz-Dmytro (changes requested):
Some bug with table, please fix comments above 🙂

src/utils/FilterSortPeople.ts
Yevheniia-Sid: > The `filter` method returns a copy of the array, so you don't need to do it manually.

That would break all my logic. I don't see how to do it without this copy.

BudnikOleksii: > > The `filter` method returns a copy of the array, so you don't need to do it manually.
> 
> That would break all my logic. I don't see how to do it without this copy.

I suggest you break your function into a few smaller ones. One function is for filtering, second is for sorting.
And pass prepared parameters, your helpers shouldn't know anything about `queryParams`, so you can use in case query, filters, and etc you get from another thing

BudnikOleksii (changes requested):
Good work 👍 
Let's refactor your `FilterSortPeople` helper as suggested, also for functions use just camelCase.
And, one moment to fix, when the user clicks on some person from the table you should keep the filter/sort parameters, check the working version

PR #519, opened 2023-10-05
Develop

cty2802:
[DEMO LINK](https://cty2802.github.io/react_people-table-advanced/)


src/components/PeopleSort/PeopleSort.tsx
```diff
@@ -0,0 +1,114 @@
...
+    if (sort) {
+      params.set('sort', sort);
+    } else {
+      params.delete('sort');
+    }
+
+    if (sort === oldSort && !oldOrder) {
+      params.set('order', 'desc');
+    } else {
+      params.delete('order');
```
polosanya, line 33: Consider using switch...case here instead

src/helper.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const humans: Person[] = [...people];
+
+  return humans.sort((humanA: Person, humanB: Person): number => {
+    if (oldSort === SortBy.Name) {
+      return oldOrder === SortOrder.Desc
+        ? (humanA.name.localeCompare(humanB.name) * -1)
+        : humanA.name.localeCompare(humanB.name);
+    }
+
+    if (oldSort === SortBy.Sex) {
```
polosanya, line 33: Same here

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,142 @@
...
+    || hasNormalizedQuery(person.motherName ?? '', query))
+    && filterByGender(person, selectedGender)
+    && filterByCentury(person, centuries)
+  ));
+}
+
+export const PeoplePage: React.FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [errorMassage, setErrorMassage] = useState('');
```
polosanya, line 49:
```suggestion
  const [errorMessage, setErrorMessage] = useState('');
```

polosanya (changes requested):
Nice! Let's improve your solution a bit)
The filters shouldn't reset when we select a person
select

src/components/PeopleSort/PeopleSort.tsx
cty2802: I considered using switch...case here, but I feel the code looks too complicated with it. The thing that I should check not only the 'sort' variable, but the oldOrder variable too. Still, I would appreciate it if you give me an example of how it can be done nicely.

src/components/PeopleSort/PeopleSort.tsx
```diff
@@ -0,0 +1,114 @@
...
+    if (columnName === oldSort && !oldOrder) {
+      return 'fas fa-sort-up';
+    }
+
+    if (columnName === oldSort && oldOrder) {
+      return 'fas fa-sort-down';
+    }
+
+    return 'fas fa-sort';
+  };
```
roman-mirzoian, line 19: Use the `classnames` library for this logic

src/components/PeopleSort/PeopleSort.tsx
```diff
@@ -0,0 +1,114 @@
...
+        <span className="is-flex is-flex-wrap-nowrap">
+          Name
+          <Link
+            to={{
+              pathname: '',
+              search: addSortToUrlParams(SortBy.Name).toString(),
+            }}
+          >
+            <span className="icon">
+              <i className={setIconClass(SortBy.Name)} />
```
roman-mirzoian, line 56: `classnames` issue

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,142 @@
...
+}
+
+function filterByGender(person: Person, selectedGender: Gender): boolean {
+  return selectedGender === Gender.All || selectedGender === person.sex;
+}
+
+function filterByCentury(person: Person, centuries: string[]):boolean {
+  return !centuries?.length
+    ? true
+    : centuries.includes(Math.floor(person.born / 100).toString());
```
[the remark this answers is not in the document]
roman-mirzoian: But avoid usign "magic values" (100)

roman-mirzoian (changes requested):
Also, if there are no search results, we have to see only the message, without the table:
image

PR #534, opened 2023-10-08
Add task solution

NikitaOrliak:
[DEMO LINK](https://NikitaOrliak.github.io/react_people-table-advanced/)

I'm sure using hook useSearchParams in the filter component is a bad idea, but it seems to me that creating constants for query, centuries and gender in the main PeoplePage component every time is much worse than what I did :)


src/helpers/GetPreparedPeople.tsx
```diff
@@ -0,0 +1,102 @@
...
+    || (person.fatherName
+      && person.fatherName.toLowerCase().includes(lowerCaseQuery)),
+  );
+};
+
+export const getPreparedPeople = (
+  people: Person[],
+) => {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const [searchParams] = useSearchParams();
```
maxim2310, line 77: you must dont use useSearchParams  here, but just pass attribute (sort, order, etc) in func

src/components/PeopleFilters.tsx
```diff
@@ -16,6 +49,16 @@ export const PeopleFilters = () => {
...
+            value={query || ''}
+            onChange={(e) => {
+              searchParams.set('query', `${e.target.value}`);
+
+              if (searchParams.get('query') === '') {
+                searchParams.delete('query');
+              }
+
+              setSearchParams(searchParams);
+            }}
```
maxim2310, line 61: need to move it to separate func

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }, []);
+
+  return (
+    <>
+      <h1 className="title">People Page</h1>
+
+      <div className="block">
+        <div className="columns is-desktop is-flex-direction-row-reverse">
+          {
+            !isError && !isLoading && !!people.length && (
```
maxim2310, line 38: need to move it to variable

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

NikitaOrliak:
I pass searchParams to the filter function. Hope it is better than pass each variable separately :)

anastasiiavorobiova (approved):
Great work!

PR #545, opened 2023-10-09
1.0

victor-buldenko:
[DEMO LINK](https://victor-buldenko.github.io/react_people-table-advanced/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve something
1. Need to remove this text
image

2. This block should be shown only after people have loaded onto the page
image

maxim2310 (changes requested):
these buttons must be clickable and change sort and order 
image

SanyaBratashchuk (approved):

PR #547, opened 2023-10-12
Develop

daryna-hnidash:
[DEMO LINK](https://daryna-hnidash.github.io/react_people-table-advanced/)


src/App.tsx
```diff
@@ -1,20 +1,50 @@
...
         </div>
       </div>
-    </div>
-  );
-};
+    </nav>
+
+    <main className="section">
+      <div className="container">
+
```
witflash, line 44: There is no need to use additional empty lines between a parent and a child.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,210 @@
...
+        'fa-sort-up': sort === sortParam
+          && !searchParams.has('order'),
+      },
+    );
+  };
+
+  const visiblePeople = sorting();
+
+  return (
+    <>
```
witflash, line 107: You can avoid using `Fragment` here. You can just return:
```
return !visiblePeople.length
  ? <p>...</p>
  : (<table>...
```
It's not a mistake, it's just a little improvement.

witflash (approved):
Wow!

I think it's excellent work!
Everything looks good both a DEMO page and your code.
Well done! 👍

PR #553, opened 2023-10-19
done

yaroslav5481:
https://yaroslav5481.github.io/react_people-table-advanced/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import { PeoplePage } from './components/PeoplePage';
+import { Outlet } from 'react-router-dom';
 import { Navbar } from './components/Navbar';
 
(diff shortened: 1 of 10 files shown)


denys-danyliuk (changes requested):
this button should be active by default
Знімок екрана 2023-10-19 о 16 33 03

when user filters list using 'Male' or 'Female' buttons names of the opposite sex should be highlighted in needed color, check the working example
Знімок екрана 2023-10-19 о 16 34 56
Знімок екрана 2023-10-19 о 16 35 04

it takes extra one click to change sorting to the other field, check how it should work
https://github.com/mate-academy/react_people-table-advanced/assets/106310796/60462063-11b9-461a-bc2b-2e092c6da77e

PR #555, opened 2023-10-22
People table advanced

Orchibald:
[DEMO LINK](https://orchibald.github.io/react_people-table-advanced/)


src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,179 @@
...
+      .catch(() => setIsError(true))
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const displayedPeople = () => {
+    let filteredPeople = [...people];
+
+    const centuries = searchParams.getAll('centuries') || [];
+    const gender = searchParams.get('sex') || '';
+    const query = searchParams.get('query') || '';
```
VitaliyBondarenko1982, line 45: suggest to make enum for all possible search params fields. and use in all places instead strings.

VitaliyBondarenko1982 (changes requested):
Good job.
Need some improvements.

After click on some person link should be save chosen filters.
Now they reseting.
Please compare working example with demo link in this moment.

PR #557, opened 2023-10-23
solution

MaciejKujawa:
[DEMO LINK](https://MaciejKujawa.github.io/react_people-table-advanced/)


choeqq (changes requested):
One minor issue - when pressing 'Reset all filters' the input isn't getting cleared, please update that and you'll be good to go ✅

PR #565, opened 2023-11-02
add task solution

ydashko:
[DEMO LINK](https://ydashko.github.io/react_people-table-advanced/)


src/components/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,104 @@
...
+      }
+
+      return { ...person, mother, father };
+    });
+
+    setPeopleFromServer(preparedPeople);
+  };
+
+  const filterPeople = useMemo(() => {
+    return filteredPeople(peopleFromServer, {
```
TarasHoliuk, line 45: just FYI: should be vice versa - const `filteredPeople` with array and function `filterPeople`. Function does some action so when I see verb in name I expect it's a function

TarasHoliuk (approved):
Very good job :heavy_check_mark:

PR #567, opened 2023-11-03
solved

silcki:
[DEMO LINK](https://silcki.github.io/react_people-table-advanced/)


anastasiiavorobiova (changes requested):
Great work! Consider showing the message instead of the table. Now the aside disappears and a user can't reset filters or apply another one 
Screenshot 2023-11-03 at 15 48 27

PR #568, opened 2023-11-03
React People Table (Advanced) - Filterging and Sorting was implementing

serge-kontakt:
[DEMO LINK](https://serge-kontakt.github.io/react_people-table-advanced/)


src/components/PeopleFilters.tsx
```diff
@@ -27,66 +67,50 @@ export const PeopleFilters = () => {
...
+                data-cy="century"
+                key={century}
+                className={classNames('button mr-1', {
+                  'is-info': centuries.includes(century),
+                })}
+                params={{
+                  centuries: centuries.includes(century)
+                    ? centuries.filter(cent => century !== cent)
+                    : [...centuries, century],
+                }}
```
VitaliyBodnar, line 81: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+    if (sort === 'born' || sort === 'died') {
+      filteredPeople = filteredPeople.sort((a, b) => a[sort] - b[sort]);
+    }
+
+    if (sort === 'name' || sort === 'sex') {
+      filteredPeople = (
+        filteredPeople.sort((a, b) => a[sort].localeCompare(b[sort]))
+      );
+    }
```
VitaliyBodnar, line 66: - it's better to use switch case statement instead of multiple if blocks 
- please also move 'born', 'died', 'name' and 'sex' to the enum as it will make a code cleaner

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,73 @@
...
+        if (peopleAll.length === 0) {
+          setIsNoPeople(true);
+        }
+      })
+      .catch(() => {
+        setIsError(true);
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
```
VitaliyBodnar, line 30:
```suggestion
      .catch(() => setIsError(true))
      .finally(() => setIsLoading(false));
```

VitaliyBodnar (changes requested):
Good job! Almost done!

SanyaBratashchuk (approved):

PR #577, opened 2023-11-09
Develop

yuliabakun:
[demo](https://yulabakun.github.io/react_people-table-advanced/)

Here is my implementation, but it is not the final version.

I am having trouble with the sorting feature. 
For now, I managed to add correct corresponding URL when clicking on sorting link,
 but I have an issue with reading the params from url and updating table correspondingly.
 Now it only sort table in asc way even if there is an order param with desc value in url.

Would appreciate some advice on how to fix that issue.


yulabakun:
Implemented sorting functionality and seems like it works pretty much as expected. 
Styles are not ready yet.

yulabakun:
Added styles for sorting icons

koros-rk (approved):
good job ;)

vasylmiskiv (approved):
GJ. Everything looks nice

PR #579, opened 2023-11-12
solution

IhorVerba:
https://IhorVerba.github.io/react_people-table-advanced/


src/stores/PeopleContext.tsx
```diff
@@ -0,0 +1,102 @@
...
+  useEffect(() => {
+    setIsLoading(true);
+
+    getPeople()
+      .then(setPeople)
+      .catch(() => setLoadingError(true))
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const filterPeople = ({
```
TarasHoliuk, line 34: For me, it's a bit inconsistent

Chose one:
- move this function into a separate file (utils.ts or smth like that) pass everything (including the people array) as an argument and use this function, import it in the `PeopleTable` component
- get all params from `useSearchParams` hook here calculate an array of filtered people and pass the prepared array as a context value

TarasHoliuk (changes requested):
GJ :+1: 
Added 2 comments to improve logic and make code more readable and clear

denys-danyliuk (changes requested):
it takes two clicks to change sorting to another field, look how it works in the working example

https://github.com/mate-academy/react_people-table-advanced/assets/106310796/7ab408e6-728e-4da5-8123-a0095a7bb108

PR #586, opened 2023-11-15
solution

Spec0s:
https://spec0s.github.io/react_people-table-advanced/

### DIFF
--- a/src/components/PeoplePage.tsx
+++ b/src/components/PeoplePage.tsx
@@ -1,34 +1,122 @@
+import { useEffect, useState } from 'react';
+import { useSearchParams } from 'react-router-dom';
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { getPeople } from '../api';
+import { Person } from '../types';
 
 export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [peopleLoader, setPeopleLoader] = useState(true);
+  const [error, setError] = useState(false);
+  const [searchParams] = useSearchParams();
+  const sex = searchParams.get('sex') || '';
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+
+  const loadPeople = async () => {
+    try {
+      setError(false);
+      setPeopleLoader(true);
+      const load = await getPeople();
+
+      setPeople(load);
+    } catch {
+      setError(true);
+    }
+
+    setPeopleLoader(false);
+  };
+
+  useEffect(() => {
+    loadPeople();
+  }, []);
+
+  const noPeopleOnServer = () => {
+    return !people.length && !error && !peopleLoader;
+  };
+
+  const peopleFiltered = () => {
+    let filtered = [...people];
+
+    if (centuries.length) {
+      filtered = [
+        ...filtered.filter(
+          person => centuries.some(
+            century => {
+              return Math.ceil(person.born / 100)
+              === Number(century);
+            },
+          ),
+        ),
+      ];
+    }
+
+    if (query) {
+      filtered = [...filtered.filter(({
+        name,
+        fatherName,
+        motherName,
+      }) => (
+        name.toLowerCase().includes(query.toLowerCase())
+        || fatherName?.toLowerCase().includes(query.toLowerCase())
+        || motherName?.toLowerCase().includes(query.toLowerCase())
+      ))];
+    }
+
+    if (sex) {
+      filtered = [...filtered.filter(person => person.sex === sex)];
+    }
+
+    return filtered;
+  };
+
   return (
     <>
       <h1 className="title">People Page</h1>
-
       <div className="block">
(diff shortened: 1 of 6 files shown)


src/components/PeoplePage.tsx
```diff
@@ -1,34 +1,122 @@
...
+      setError(false);
+      setPeopleLoader(true);
+      const load = await getPeople();
+
+      setPeople(load);
+    } catch {
+      setError(true);
+    }
+
+    setPeopleLoader(false);
```
anastasiia-tilikina, line 29: this is good to put in finally branch

src/components/PeoplePage.tsx
```diff
@@ -1,34 +1,122 @@
...
+
+    if (query) {
+      filtered = [...filtered.filter(({
+        name,
+        fatherName,
+        motherName,
+      }) => (
+        name.toLowerCase().includes(query.toLowerCase())
+        || fatherName?.toLowerCase().includes(query.toLowerCase())
+        || motherName?.toLowerCase().includes(query.toLowerCase())
```
anastasiia-tilikina, line 64: move this check to separate helper to avoid code duplicates, like this:

```typescript
const isMatch = (str: string) => str.toLowerCase().includes(query.toLowerCase())
```

src/components/PeoplePage.tsx
```diff
@@ -1,34 +1,122 @@
...
+    }
+
+    setPeopleLoader(false);
+  };
+
+  useEffect(() => {
+    loadPeople();
+  }, []);
+
+  const noPeopleOnServer = () => {
```
anastasiia-tilikina, line 36: this could be not a fn but just variable

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,171 @@
...
-          <td>1728</td>
-          <td>1734</td>
-          <td>Livina de Vrieze</td>
-          <td>
-            <a href="#/people/pieter-bernard-haverbeke-1695">
-              Pieter Bernard Haverbeke
-            </a>
-          </td>
-        </tr>
+  const sortedPeople = () => {
```
anastasiia-tilikina, line 26: it's not a good practice to put true as switch condition.
you can use sort value instead.

and consider using [destructuring assignment](https://javascript.info/destructuring-assignment#array-destructuring) for swapping variables if there is desc order. lide this:

```typescript
let [personA, personB] = [a, b];

if (order === 'desc') {
  [personA, personB] = [personB, personA]
}
```

anastasiia-tilikina: and don't forget to make a copy since sort is mutating method

anastasiia-tilikina: and you can put switch case inside sort method to avoid code duplicates

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,171 @@
...
+                  to={`/people/${personLink(fatherName)}`}
+                >
+                  {fatherName || '-'}
+                </Link>
+              </td>
+            ) : (
+              <td>
+                {fatherName || (
+                  '-'
+                )}
```
anastasiia-tilikina, line 164:
```suggestion
                {fatherName ||  '-'}
```

no need to write it from a new line

anastasiia-tilikina (changes requested):
Very well done!

let's just fix a couple of issues

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,164 @@
...
-          <td>1734</td>
-          <td>Livina de Vrieze</td>
-          <td>
-            <a href="#/people/pieter-bernard-haverbeke-1695">
-              Pieter Bernard Haverbeke
-            </a>
-          </td>
-        </tr>
+  const sortedPeople = () => {
+    let sorted = [];
```
anastasiia-tilikina, line 27:
```suggestion
```

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,164 @@
...
-              href="#/people/livina-haverbeke-1692"
-            >
-              Livina Haverbeke
-            </a>
-          </td>
-          <td>
-            <a href="#/people/pieter-de-decker-1705">Pieter de Decker</a>
-          </td>
-        </tr>
+    sorted = [...people.sort((a, b) => {
```
anastasiia-tilikina, line 29:
```suggestion
   const sorted = [...people.sort((a, b) => {
```
you can create var like this. no need to assign an empty arr first

src/components/PeopleTable.tsx
```diff
@@ -1,682 +1,164 @@
...
+                  className="has-text-danger"
+                >
+                  {motherName || '-'}
+                </Link>
+              </td>
+            ) : (
+              <td>
+                {motherName || (
+                  '-'
+                )}
```
anastasiia-tilikina, line 144:
```suggestion
                {motherName ||  '-' }
```

be consistent and fix everywhere

anastasiia-tilikina (approved):
Well done!

It's approve, but pay attention to my comments!

PR #599, opened 2023-11-20
add task solution

dsfreedom:
[DEMO LINK](https://dsfreedom.github.io/react_people-table-advanced/)


BetterSol (approved):
Looks great 🔥 
Great idea to create a Page Folder! 
Suggested a few minor changes to improve your code even more;)

src/utils/filterFunction.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const getVisiblePeople = (
+  recievedPeople: Person[],
+  params: URLSearchParams,
+) => {
+  const sex = params.get('sex') || '';
+  const query = params.get('query') || '';
+  const centuries = params.getAll('centuries') || [];
+  const sort = params.get('sort') || '';
+  const order = params.get('order') || '';
```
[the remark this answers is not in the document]
dsfreedom: Somehow if I remove those the test 'should allow to choose a person' fails, so I've kept it, but improved the code with your other suggestions, thank you a lot!)

src/Root.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <Route path="/" element={<App />}>
+        <Route path="/" element={<HomePage />} />
+
+        <Route path="/home" element={<Navigate to="/" replace />} />
+
+        <Route path="/people" element={<PeoplePage />}>
+          <Route
+            path="/people/:slug?"
+            element={<PeoplePage />}
+          />
```
superpooperxxx, line 24: As this Route is a nested route of `/people` you don't have to specify `/people/:slug?`, just `:slug?`

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,64 @@
...
+        >
+          Male
+        </SearchLink>
+
+        <SearchLink
+          params={{ sex: 'f' }}
+          className={classNames({ 'is-active': sex === 'f' })}
+        >
+          Female
+        </SearchLink>
```
superpooperxxx, line 61: Render as a list

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,64 @@
...
+    const index = updatedCenturies.indexOf(`${century}`);
+
+    if (updatedCenturies.includes(century)) {
+      updatedCenturies.splice(index, 1);
+    } else {
+      updatedCenturies.push(century);
+    }
+
+    return { centuries: updatedCenturies };
+  };
```
superpooperxxx, line 35: Why not doing as Misha showed in video ?
```
const updatedCenturies = currentCenturies.includes(century)
  ? currentCenturies.filter()
  : [...currentCenturies, century]

File: src/components/PeopleFilters.tsx [hunk 5]
```diff
@@ -27,66 +81,40 @@ export const PeopleFilters = () => {
...
-            </a>
+            </SearchLink>
           </div>
         </div>
       </div>
 
       <div className="panel-block">
-        <a
+        <SearchLink
+          params={{ sex: null, query: null, centuries: null }}
```
superpooperxxx, line 113: What if you have 10 filters? Would set `null` to all of them? What if you do `params={new URLSearchParams()}`?

superpooperxxx (changes requested):
Almost a genius solution🔥

src/components/PeopleFilters.tsx
dsfreedom: Out of interest, if we had to reset only filters and not sorting params (that's what I thought had to be done), how would you write it?

superpooperxxx: Oh actually I would create a new URLSearchParams object and I would add there fields that should remain

superpooperxxx (approved):
🔥

PR #600, opened 2023-11-20
sol1

A1eeex:
[DEMO LINK](https://A1eeex.github.io/react_people-table-advanced/)


adioskv (changes requested):
### Good job! 🚀
Only 1 mistake, when you choose a person, your search parameters are dissapearing.
### Example:

Before choosing a person:
image

After:
image

### Expected behaviour

Before: 
image

After:
image

src/components/PeopleFilters.tsx
```diff
@@ -1,17 +1,70 @@
+// import { useState } from 'react';
+import cn from 'classnames';
+import { useSearchParams } from 'react-router-dom';
+import { SearchLink } from './SearchLink';
+import { Sex } from '../types/Sex';
+
 export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  // const { pathname, search } = useLocation();
```
vasylmiskiv, line 9: should remove location if it's not needed

src/components/PeopleFilters.tsx
```diff
@@ -27,66 +80,52 @@ export const PeopleFilters = () => {
...
-              20
-            </a>
+            {centurysArr.map((century) => {
+              return (
+                <SearchLink
+                  key={century}
+                  params={{
+                    century: centuries.includes(century)
+                      ? centuries.filter((c) => c !== century)
+                      : [...centuries, century],
```
vasylmiskiv, line 90: Would be better to move the logic to separate function and pass the century property

vasylmiskiv (approved):

PR #602, opened 2023-11-20
solution

adioskv:
[DEMO LINK](https://adioskv.github.io/react_people-table-advanced/)


Ayogoharo:
If chose sort sort type and make sort order desc and then chose other sort type, it will not be chosen.
https://www.loom.com/share/d2d0c55fb94346f8a89c5a463f7c1ceb

Ayogoharo:
If there are no people with the same name \ motherName \ fatherName as in query should give error message

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,100 @@
...
+                      ? (
+                        <p data-cy="noPeopleMessage">
+                          There are no people on the server
+                        </p>
+                      )
+                      : <PeopleTable people={filteredPeople} />}
+                  </>
+                )}
+            </div>
+          </div>
```
Ayogoharo, line 95:
```jsx
              {(() => {
                switch (true) {
                  case isLoading:
                    return (<Loader />);
                  case error:
                    return (
                      <p
                        data-cy="peopleLoadingError"
                        className="has-text-danger"
                      >
                        Something went wrong
                      </p>
                    );
                  case !peopleFromServer.length:
                    return (
                      <p data-cy="noPeopleMessage">
                        There are no people on the server
                      </p>
                    );
                  case !!peopleFromServer.length:
                    return (<PeopleTable people={visiblePeople} />);
                  default:
                    return (<></>);
                }
              })()}
            </div>
            ```

Comment (Ayogoharo) [reply src/pages/PeoplePage/PeoplePage.tsx:95]: You can do switch instead of 2  "if else"

**Review by Ayogoharo — APPROVED** [review]
looks good to me, just need to fix two little mistakes

File: src/pages/PeoplePage/PeoplePage.tsx [hunk 2]
```diff
@@ -0,0 +1,100 @@
...
+import { sortPeople } from '../../utils/sortPeople';
+
+export const PeoplePage = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [people, setPeople] = useState<Person[] | []>([]);
+  const [error, setError] = useState(false);
+
+  const [searchParams] = useSearchParams();
+
+  const filters: Filters = useMemo(() => {
```
vasylmiskiv, line 20: 👍

vasylmiskiv (approved):
lgtm

PR #604, opened 2023-11-21
better late than never

HardCodWorker:
[DEMO LINK](https://HardCodWorker.github.io/react_people-table-advanced/)


src/components/Navbar.tsx
```diff
@@ -1,24 +0,0 @@
-export const Navbar = () => {
```
dsfreedom, line 1: You've had this file for a reason))))

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,150 @@
+import { Link, useSearchParams } from 'react-router-dom';
+import classNames from 'classnames';
+
+export const PeopleFilters = () => {
+  const centuriesList = [16, 17, 18, 19, 20];
```
dsfreedom, line 5: I suggest writing it before you write functional component and in all capitals as it is a constant:
const CENTURIES_LIST = [16, 17, 18, 19, 20];

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    params.delete('centuries');
+    if (century !== '') {
+      newCenturies.forEach(cen => params.append('centuries', cen.toString()));
+    }
+
+    return params.toString();
+  };
+
+  const clear = () => {
```
dsfreedom, line 51: I think it is better to name this function a more descriptive name, rather than clear. As for example handleResetAll or onReset 🤷‍♂️

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const params = new URLSearchParams(searchParams);
+
+    params.delete('sex');
```
dsfreedom, line 54: Also, I understand trying to get it in your own way, but as a suggestion: we have a SearchLink file and a helper function in utilities (getSearchWith) and they are already connected, so you could've used that and write much less (you would only need to pass sex: null, query: null ... in params for SearchLink. Again, it's only a suggestion of how it could be done in another way. Also this way you would not need to write handle centuries and handleSex

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+  const displayedPeople = getfilteredPeople();
+
+  return (
+    <>
+      <h1 className="title">People Page</h1>
+      <div className="block">
+        <div className="columns is-desktop is-flex-direction-row-reverse">
+          <div className="column is-7-tablet is-narrow-desktop">
```
dsfreedom, line 91: When the people are loaded we see a 0:
image

The issue with this I'm pretty sure is hidden here as this is how conditional operator && works (it returns zero if the condition is false. So for this to not happen you can use ternary operators or create a separate function (function would be the most efficient way here as you also created components for each version of a page display))

dsfreedom (approved):
Great work!!, but please solve the issue with zero)

src/components/PeopleFilters/PeopleFilters.tsx
HardCodWorker: Yeah I know,  just did not want to use a premaid solution, in order to train

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,150 @@
...
+import classNames from 'classnames';
+
+export const PeopleFilters = () => {
+  const centuriesList = [16, 17, 18, 19, 20];
+  const [searchParams, setSearchParams] = useSearchParams();
+  const query = searchParams.get('query') || '';
+  const sex = searchParams.get('sex') || 'all';
+  const centuries = searchParams.getAll('centuries') || [];
+
+  const handleQuery = (ch: string) => {
```
vasylmiskiv, line 11: You should write more appropriate name of the parameter instead of `ch`

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const handleCenturies = (century: string) => {
+    const params = new URLSearchParams(searchParams);
+
+    const newCenturies = centuries.includes(century.toString())
+      ? centuries.filter(cent => cent !== century.toString())
+      : [...centuries, century];
+
+    params.delete('centuries');
+    if (century !== '') {
```
vasylmiskiv, line 44:
```suggestion
    if (century) {
```

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,19 @@
+import { Link } from 'react-router-dom';
+import classNames from 'classnames';
+import { Person } from '../../types';
+
+interface Props {
+  person: Person | undefined
```
vasylmiskiv, line 6: It will be the same
```suggestion
  person?: Person;
```

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,171 @@
+import { Link, useParams, useSearchParams } from 'react-router-dom';
+import cn from 'classnames';
+import { Person } from '../../types';
+import { PersonLink } from '../PersonLink/PersonLink';
+
+interface Props {
+
+  people: Person[] | undefined
```
vasylmiskiv, line 8:
```suggestion
  people?: Person[];
```

vasylmiskiv (approved):
Overall looks fine but pay attention to the comments

PR #609, opened 2023-11-21
Develop

denzzlle171:
[DEMO LINK](https://denzzlle171.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -10,8 +10,6 @@ export const PeopleFilters = () => {
   const query = searchParams.get('query') || '';
   const centuries = searchParams.getAll('centuries') || [];
 
-  // console.log(centuries);
-
   const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
     const newQueryParam = getSearchWith(searchParams, {
       query: event.target.value.trim().toLocaleLowerCase() || null,
@@ -69,7 +67,7 @@ export const PeopleFilters = () => {
       <div className="panel-block">
         <div className="level is-flex-grow-1 is-mobile" data-cy="CenturyFilter">
           <div className="level-left">
-            {centuriesItem.map(century => {
+            {centuriesItem.map((century) => {
               return (
                 <SearchLink
                   key={century}
@@ -89,21 +87,28 @@ export const PeopleFilters = () => {
           </div>
 
           <div className="level-right ml-4">
-            <a
+            <SearchLink
               data-cy="centuryALL"
-              className="button is-success is-outlined"
-              href="#/people"
+              className={classnames('button', 'is-success', {
+                'is-outlined': centuries.length > 0,
+              })}
+              params={{ centuries: null }}
+              // to="#/people"
             >
               All
-            </a>
+            </SearchLink>
           </div>
         </div>
       </div>
 
       <div className="panel-block">
-        <a className="button is-link is-outlined is-fullwidth" href="#/people">
+        <SearchLink
+          className="button is-link is-outlined is-fullwidth"
+          // to="#/people"
+          params={{ centuries: null, query: null, sex: null }}
+        >
           Reset all filters
-        </a>
+        </SearchLink>
       </div>
     </nav>
   );
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,12 +1,52 @@
+import classnames from 'classnames';
+import { useSearchParams } from 'react-router-dom';
+import { SearchLink } from './SearchLink';
+import { getSearchWith } from '../utils/searchHelper';
+
 export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const sex = searchParams.get('sex');
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+
+  // console.log(centuries);
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const newQueryParam = getSearchWith(searchParams, {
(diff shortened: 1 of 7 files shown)


src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,107 @@
...
+    setIsLoad(true);
+    getPeople()
+      .then((peopleFromServer) => setPeople(peopleFromServer))
+      .catch(() => {
+        setError('Something went wrong');
+      })
+      .finally(() => setIsLoad(false));
+  }, []);
+
+  const getVisiblePeople = () => {
```
dolphinnq, line 32: Better move this function outside the component and just call it here with your arguments.

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,107 @@
...
+  const query = searchParams.get('query')?.trim().toLocaleLowerCase();
+  const centuries = searchParams.getAll('centuries') || [];
+  const sort = searchParams.get('sort');
+  const order = searchParams.get('order');
+
+  useEffect(() => {
+    setIsLoad(true);
+    getPeople()
+      .then((peopleFromServer) => setPeople(peopleFromServer))
+      .catch(() => {
```
dolphinnq, line 26: Try to catch an `error` here, maybe it will fix some of your failed tests

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,107 @@
...
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { getPeople } from '../api';
+
+import { Person } from '../types/Person';
 
 export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[] | null>(null);
+  const [isLoad, setIsLoad] = useState<boolean>(false);
```
dolphinnq, line 12: It is better to make `isLoad` `true` by default. Now there's no need to set it `true` in line `23`

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,107 @@
...
-
-              <PeopleTable />
+              {isLoad && <Loader />}
+              {error && (
+                <p data-cy="peopleLoadingError" className="has-text-danger">
+                  {error}
+                </p>
+              )}
+              {people && <PeopleTable people={getVisiblePeople()} />}
+              {!isLoad && !error && !people && (
```
dolphinnq, line 102: `no people` appears before page is loaded for a moment. Create a variable for this and try to change your `people` condition. 
image

dolphinnq (approved):
Good job!

src/components/PeoplePage.tsx
denzzlle171: Maybe I've already fixed it with lowder, because I don't see this problem now

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,66 @@
...
+
+  const sex = searchParams.get('sex');
+  const query = searchParams.get('query')?.trim().toLocaleLowerCase();
+  const centuries = searchParams.getAll('centuries') || [];
+  const sort = searchParams.get('sort');
+  const order = searchParams.get('order');
+
+  useEffect(() => {
+    getPeople()
+      .then((peopleFromServer) => setPeople(peopleFromServer))
```
superpooperxxx, line 24: `.then(setPeople)`

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,66 @@
...
+import { Person } from '../types/Person';
 
 export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[] | null>(null);
+  const [isLoad, setIsLoad] = useState<boolean>(true);
+  const [error, setError] = useState<string | null>(null);
+  const [searchParams] = useSearchParams();
+
+  const sex = searchParams.get('sex');
+  const query = searchParams.get('query')?.trim().toLocaleLowerCase();
```
superpooperxxx, line 17: Not sure why you need to trim and make lower cased

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,66 @@
...
+              {people && (
+                <PeopleTable
+                  people={getVisiblePeople(
+                    people,
+                    sex,
+                    query,
+                    centuries,
+                    sort,
+                    order,
+                  )}
```
superpooperxxx, line 58: 1. For better readability store in separate variable
2. Just pass the whole `searchParams` object

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,50 @@
...
+        >
+          Male
+        </SearchLink>
+
+        <SearchLink
+          className={classnames({ 'is-active': sex === 'f' })}
+          params={{ sex: 'f' }}
+        >
+          Female
+        </SearchLink>
```
superpooperxxx, line 47: Render as list

src/components/PeopleFilters.tsx
```diff
@@ -27,66 +67,46 @@ export const PeopleFilters = () => {
...
           </div>
         </div>
       </div>
 
       <div className="panel-block">
-        <a
+        <SearchLink
           className="button is-link is-outlined is-fullwidth"
-          href="#/people"
+          params={{ centuries: null, query: null, sex: null }}
```
superpooperxxx, line 106: Find a better way to reset all by just 1 line of code. If you had 20 filters, would you set null to all of them?

superpooperxxx (changes requested):
Almost there🔥

src/components/PeopleFilters.tsx
denzzlle171: all the other options I can think of
will either just increase the amount of code or reset the sorting as well.

superpooperxxx (approved):
🔥

PR #614, opened 2023-11-21
solution

vlevik:
[DEMO LINK](https://vlevik.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeoplePage.tsx
+++ b/src/components/PeoplePage.tsx
@@ -1,8 +1,79 @@
+import { useEffect, useState } from 'react';
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { Person } from '../types';
+import { getPeople } from '../api';
 
 export const PeoplePage = () => {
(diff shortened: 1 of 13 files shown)


src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,97 @@
+import { useEffect, useState } from 'react';
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { Person } from '../types';
+import { getPeople } from '../api';
 
 export const PeoplePage = () => {
+  const [filteredPeople, setFilteredPeople] = useState<Person[]>([]);
+  const [allPeople, setAllPeople] = useState<Person[]>([]);
```
Kuper18, line 10: is it necessary to use two the same states with the same data

Kuper18 (changes requested):
Overall nice work! But I noticed one issue in your app, when you set some filter parameters, and when you click on some person to highlight the row, the filter parameters are reset to the default url, but they shouldn't, check more carefully demo version

superpooperxxx (approved):
🔥

PR #623, opened 2023-11-23
Develop

Stim1:
[DEMO LINK](https://Stim1.github.io/react_people-table-advanced/)


src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [selectedPersonSlug, setSelectedPersonSlug] = useState<string>('');
+  const [filterGenderStatus, setFilterGenderStatus]
+  = useState<Gender>(Gender.ALL);
+  const [filteredPeople, setFilteredPeople] = useState<Person[]>([]);
+  const [selectedCenturies, setSelectedCenturies] = useState<string[]>([]);
+  const [searchQuery, setSearchQuery] = useState('');
+  const sort = searchParams.get('sort');
+  const order = searchParams.get('order');
+  const { slug } = useParams();
```
superpooperxxx, line 27: I think this toooo much code for one component. Let's do following:
1. Let's put filtration to separate function in separate file instead of useEffect. This function should take `searchParams` object. Store the result of filtration in variable.
2. `query` handling should be in `PeopleFilters`

Why do you need all those states like `selectedPersonSlug`, `filterGenderStatus`, etc. This the whole point of using router, because your url parts are basically a state.

superpooperxxx (changes requested):
Hey, I think you slightly don't understand the usage of router. I suggest watching sections closely again.

src/pages/PeoplePage.tsx
Stim1: yes, this topic is quite difficult for me, I realized the mistakes and will redo this task

PR #624, opened 2023-11-23
develop

atrofymovych:
[DEMO LINK](https://atrofymovych.github.io/react_people-table-advanced/)


src/App.tsx
```diff
@@ -1,20 +1,29 @@
...
-      <div className="section">
+      <main className="section">
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
+            <Route path="/" element={<HomePage />} />
+            <Route path="/people" element={<PeoplePage />}>
+              <Route path=":personSlug?" />
```
TarasHoliuk, line 17:
```suggestion
              <Route path=":personSlug" />
```

src/PeopleContext.tsx
```diff
@@ -0,0 +1,133 @@
...
+export const PeopleContent: React.FC<Props> = ({ children }) => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isPageActive, setIsPageActive] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  const [searchParams, setSearchParams] = useSearchParams();
+  const sex = searchParams.get('sex') || '';
+  const centuriesUrl = searchParams.getAll('centuries') || [];
+  const query = searchParams.get('query') || '';
```
TarasHoliuk, line 71: doesn't make sense to save all these values in the context
just get them using `useSearchParams` hook

That also means you don't need context at all. For example, the `filterPeople` function is needed only in `PeoplePage` component, so it's better to define it there. The same with other values.

Because in result you've got a big complex Context component with all logic

src/components/PeoplePage.tsx
```diff
@@ -15,17 +19,46 @@ export const PeoplePage = () => {
...
+                      <thead>
+                        <tr>
+                          <th>Name</th>
+                          <th>Sex</th>
+                          <th>Born</th>
+                          <th>Died</th>
+                          <th>Mother</th>
+                          <th>Father</th>
+                        </tr>
+                      </thead>
```
TarasHoliuk, line 55: Move it inside the `PeopleTable` component - it **is** a part of table

TarasHoliuk (changes requested):
Hey! Everything works good :+1: 
But let's do some refactoring to improve your code (see coments)

src/PeopleContext.tsx
```diff
@@ -0,0 +1,59 @@
...
+    getPeople()
+      .then(usersFromServer => {
+        setPeople(usersFromServer);
+      })
+      .catch(() => {
+        setError('Something went wrong');
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
```
VitaliyBodnar, line 44: can be less
```suggestion
      .then(setPeople)
      .catch(() => setError('Something went wrong'))
      .finally(() =>setIsLoading(false));
```

src/components/PeoplePage.tsx
```diff
@@ -15,17 +19,24 @@ export const PeoplePage = () => {
...
+                {!isLoading && people.length === 0 && (
+                  <p data-cy="noPeopleMessage">
+                    There are no people on the server
+                  </p>
+                )}
+                {isLoading && !error ? <Loader /> : (
+                  <>
+                    <PeopleTable />
+                  </>
+                )}
```
VitaliyBodnar, line 38: redundant 
```suggestion
                {isLoading && !error ? <Loader /> : (
                  <PeopleTable />
                )}
```

src/components/PeoplePage.tsx
```diff
@@ -15,17 +19,24 @@ export const PeoplePage = () => {
...
-              <p>There are no people matching the current search criteria</p>
-
-              <PeopleTable />
+              <div className="box table-container">
+                {error
+                  && (
+                    <p data-cy="peopleLoadingError" className="has-text-danger">
+                      Something went wrong
+                    </p>
+                  )}
```
VitaliyBodnar, line 28:
```suggestion
               {error && (
                  <p data-cy="peopleLoadingError" className="has-text-danger">
                    Something went wrong
                  </p>
                )}
```

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,70 @@
...
+          className={cn('', { 'has-text-danger': person.sex === 'f' })}
+        >
+          {person.name}
+        </a>
+      </td>
+
+      <td>{person.sex}</td>
+      <td>{person.born}</td>
+      <td>{person.died}</td>
+      {findPersonMother(people, person)
```
VitaliyBodnar, line 47: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

VitaliyBodnar (changes requested):

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,73 @@
...
+        ? (
+          <td>
+            <a
+              href={`#/people/${motherSlug}`}
+              className="has-text-danger"
+            >
+              {person.motherName}
```
loralevitska, line 57: Destructure `person`

loralevitska (changes requested):
Almost perfect) Let's improve

- remove redundant second background
image

anastasiiavorobiova (approved):
Great work!

PR #632, opened 2023-11-29
impl

Daniilart01:
[DEMO LINK](https://daniilart01.github.io/react_people-table-advanced/)


src/components/PeopleFilters.tsx
```diff
@@ -16,6 +48,11 @@ export const PeopleFilters = () => {
             type="search"
             className="input"
             placeholder="Search"
+            value={searchParams.get('query') || ''}
+            onChange={e => setSearchParams(getSearchWith(
+              searchParams,
+              { query: e.target.value || null },
+            ))}
```
loralevitska, line 55: Move to separate function  (transfer the function to a variable, and here only write the name of the function)

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const getSortingParams = useCallback((param: string): SearchParams => {
+    if (searchParams.get('sort') === param
+      && searchParams.get('order') === 'desc') {
+      return { sort: null, order: null };
+    }
+
+    if (searchParams.get('sort') === param) {
+      return { sort: param, order: 'desc' };
+    }
```
loralevitska, line 26: Use [`switch case`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) in cases where you constantly compare one value with others

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Loader } from '../../components/Loader';
+import { getPeople } from '../../api';
+import { Person } from '../../types';
+import { PeopleFilters } from '../../components/PeopleFilters';
+import { PeopleTable } from '../../components/PeopleTable';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [error, setError] = useState(false);
+  const [loading, setLoading] = useState(false);
```
loralevitska, line 15: Rename all boolean variables according to [tips](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig)
```suggestion
  const [isError, setIsError] = useState(false);
  const [isLoading, setIsLoading] = useState(false);
```

src/components/PeopleTable/PeopleTable.tsx
Daniilart01: Your suggestion makes code harder to understand
```
switch (searchParams.get('sort')) {
      case param:
        if (searchParams.get('order') === 'desc') {
          return { sort: null, order: null };
        }

        return { sort: param, order: 'desc' };
      default: return { sort: param, order: null };
    }
```
I`d preffer to leave my code as it appears

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,136 @@
...
+            data-cy="person"
+            className={cn(
+              { 'has-background-warning': person.slug === humanId },
+            )}
+          >
+            <td>
+              <PersonLink person={person} />
+            </td>
+
+            <td>{person.sex}</td>
```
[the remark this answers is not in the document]
Daniilart01: Here I need `person` as a simple object to pass it into `PersonLink` and in `PersonLink` there is no reason to destructure person

loralevitska (approved):
You still have a lot of code duplicate, fix them 
It is possible to destructure and use the object at the same time
image
image

Daniilart01:
> You still have a lot of code duplicate, fix them It is possible to destructure and use the object at the same time image image

got u, fixed this

PR #639, opened 2023-12-05
table_advansed

MaksymKos:
https://maksymkos.github.io/react_people-table-advanced/


src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,50 @@
...
+import { SearchLink } from './SearchLink';
+import { getSearchWith } from '../utils/searchHelper';
+
 export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const centuriesList = ['16', '17', '18', '19', '20'];
+
+  const sex = searchParams.get('sex');
+  const centuries = searchParams.getAll('centuries');
+  const query = searchParams.get('query') || '';
```
VitaliyBondarenko1982, line 13: suggest to make enum for all possible searchParams

VitaliyBondarenko1982 (changes requested):
Looks good.
Compare your implementation with working example.
А couple mismatches founded.
- when exists some searchParams and click on person link - searchParams shouldn't be resetted
- when order in desc direction- should be visible only bottom arrow 
actual:
actual
working example:
working_example

src/components/PeoplePage.tsx
```diff
@@ -1,12 +1,102 @@
...
+        {
+          prepared.sort((a, b) => a.name.localeCompare(b.name));
+          break;
+        }
+
+        case 'sex':
+        {
+          prepared.sort((a, b) => a.sex.localeCompare(b.sex));
+          break;
+        }
```
etojeDenys, line 74:
```suggestion
        case 'name':
        case 'sex':
        {
          prepared.sort((a, b) => a[sort].localeCompare(b[sort]));
          break;
        }
```
you can group some cases

etojeDenys (changes requested):
i cannot find PeopleTable component in github

VitaliyBondarenko1982 (approved):
Well done.

PR #640, opened 2023-12-05
🌵

Masha-Cactus:
[DEMO LINK](https://<your_account>.github.io/react_people-table-advanced/)


src/components/Navbar.tsx
```diff
@@ -1,24 +1,31 @@
...
+        <NavLink className={getLinkNavClass} to="/">
+          Home
+        </NavLink>
+
+        <NavLink
+          className={getLinkNavClass}
+          to="/people"
+        >
+          People
+        </NavLink>
```
Viktor-Kost, line 27: it's better to move nav links data to an array an render as array

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,61 @@
...
+import { Filter } from '../types/Filter';
+import { getSearchWith } from '../utils/searchHelper';
+
+export const PeopleFilters: React.FC = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const sex = searchParams.get(Filter.SEX) || null;
+  const name = searchParams.get(Filter.NAME) || '';
+  const centuries = searchParams.getAll('centuries') || null;
+
+  const centuriesValues = [16, 17, 18, 19, 20];
```
Viktor-Kost, line 14: move this varible outside the component since it has no component dependencies

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,61 @@
...
+import classNames from 'classnames';
+import { Link, useSearchParams } from 'react-router-dom';
+import { Filter } from '../types/Filter';
+import { getSearchWith } from '../utils/searchHelper';
+
+export const PeopleFilters: React.FC = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const sex = searchParams.get(Filter.SEX) || null;
+  const name = searchParams.get(Filter.NAME) || '';
+  const centuries = searchParams.getAll('centuries') || null;
```
Viktor-Kost, line 12: why do you need to set false value as fallback? you will get falsy value anyway if you don't find a param
```suggestion
  const sex = searchParams.get(Filter.SEX);
  const name = searchParams.get(Filter.NAME);
  const centuries = searchParams.getAll('centuries');
```

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,61 @@
...
+  const centuriesValues = [16, 17, 18, 19, 20];
+
+  const resetParams = () => {
+    const newSearchParams = new URLSearchParams('');
+
+    return newSearchParams.toString();
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  function setParams(params: any) {
```
Viktor-Kost, line 23: using `any` is not a good practice. Specify what exactly you will get (or write a generic)

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,61 @@
...
+    const search = getSearchWith(params, searchParams);
+
+    setSearchParams(search);
+  }
+
+  function handleQueryChange(e: React.ChangeEvent<HTMLInputElement>) {
+    setParams({ query: e.currentTarget.value || null });
```
Viktor-Kost, line 30: guess null is redundant

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,61 @@
...
   return (
     <nav className="panel">
       <p className="panel-heading">Filters</p>
 
       <p className="panel-tabs" data-cy="SexFilter">
-        <a className="is-active" href="#/people">All</a>
-        <a className="" href="#/people?sex=m">Male</a>
-        <a className="" href="#/people?sex=f">Female</a>
+        <Link
+          className={classNames({ 'is-active': sex === null })}
```
Viktor-Kost, line 40: why are you comparing with null? you can just check does sex param exist. If it doesn't you use `all` by default (or even don't use this param)

src/components/PeopleFilters.tsx
```diff
@@ -27,66 +78,46 @@ export const PeopleFilters = () => {
...
+                key={number}
+                className={classNames('button mr-1', {
+                  'is-info': centuries?.includes(String(number)),
+                })}
+                to={{
+                  search: getSearchWith({
+                    centuries: centuries.includes(String(number))
+                      ? centuries.filter(num => String(number) !== num)
+                      : [...centuries, String(number)],
+                  }, searchParams),
```
Viktor-Kost, line 94: move this logic to a variable.
Also you can simplify this logic since know you go through the array twice (includes and filter). So, you can just use filter or something

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { Filter } from '../types/Filter';
+import { ORDER_PARAM, SORT_PARAM, Sort } from '../types/Sort';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<PersonType[]>([]);
+  const [isError, setIsError] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+  const [searchParams] = useSearchParams();
+
+  function loadPeople() {
```
Viktor-Kost, line 17: check this case. You haven't load people yet but we can see a notification that people are not found
Screenshot 2023-12-06 at 19 33 42

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    if (name) {
+      filteredPeople = filteredPeople.filter(person => person.name.toLowerCase()
+        .includes(name.toLowerCase())
+        || person.fatherName?.toLowerCase().includes(name.toLowerCase())
+        || person.motherName?.includes(name.toLowerCase()));
+    }
+
+    if (centuries) {
```
Viktor-Kost, line 67: you iterate with array so many times. 
You filter filtered values that were filtered before. It's not good. 
You should filter an array as less as possible. Try to make it only once. And in the process you filter you can add conditions inside how to filter

Viktor-Kost (changes requested):

denys-danyliuk (changes requested):
this point from requirements is not implemented
Знімок екрана 2023-12-21 о 14 53 38

there should be button looking down when desc sorting is selected
Знімок екрана 2023-12-21 о 14 55 31

src/components/Navbar.tsx
```diff
@@ -1,24 +1,32 @@
...
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">Home</a>
+import classNames from 'classnames';
+import { NavLink } from 'react-router-dom';
 
-          <a
-            aria-current="page"
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
+const links = ['Home', 'People'];
```
etojeDenys, line 4:
```suggestion
const LINKS = ['Home', 'People'];
```
we usually name constants in uppercase

etojeDenys (approved):
looks good for me

PR #667, opened 2023-12-23
solution

YuriiHlushenko:
[DEMO LINK](https://YuriiHlushenko.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,12 +1,53 @@
+import { useSearchParams } from 'react-router-dom';
+import cn from 'classnames';
+import { getSearchWith } from '../utils/searchHelper';
+import { SearchLink } from './SearchLink';
+
 export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const query = searchParams.get('query') || '';
+  const sex = searchParams.get('sex') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+
+  function setSearchWith(params: any) {
+    const search = getSearchWith(searchParams, params);
+
+    setSearchParams(search);
+  }
+
+  function handleQueryChange(event: React.ChangeEvent<HTMLInputElement>) {
+    setSearchWith({ query: event.target.value || null });
+  }
+
   return (
     <nav className="panel">
       <p className="panel-heading">Filters</p>
 
       <p className="panel-tabs" data-cy="SexFilter">
-        <a className="is-active" href="#/people">All</a>
-        <a className="" href="#/people?sex=m">Male</a>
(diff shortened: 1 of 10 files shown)


src/components/PeopleFilters.tsx
```diff
@@ -27,66 +70,49 @@ export const PeopleFilters = () => {
...
-            </a>
-
-            <a
-              data-cy="century"
-              className="button mr-1"
-              href="#/people?centuries=20"
-            >
-              20
-            </a>
+            {['16', '17', '18', '19', '20'].map((century: string) => (
```
etojeDenys, line 73: better to move this array to some constant out of the component

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,173 @@
...
+
+      if (centuries.includes('18')) {
+        peopleOfCentyries = [...peopleOfCentyries,
+          ...result.filter(
+            (person: Person) => Math.ceil(person.born / 100) === 18,
+          )];
+      }
+
+      if (centuries.includes('19')) {
+        peopleOfCentyries = [...peopleOfCentyries,
```
etojeDenys, line 92: try to simplify this logic using a loop

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,173 @@
...
+            return p1.sex.localeCompare(p2.sex);
+
+          case 'born':
+            return p1.born - p2.born;
+
+          case 'died':
+            return p1.died - p2.died;
+
+          default:
+            return 1;
```
etojeDenys, line 124:
```suggestion
        switch (sort) {
          case 'name':
          case 'sex':
            return p1[sort].localeCompare(p2[sort]);

          case 'born':
          case 'died':
            return p1[sort] - p2[sort];

          default:
            return 1;
```

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,84 @@
...
+            >
+              {motherName}
+            </a>
+          )
+          : motherName
+      );
+    }
+
+    return '-';
+  }
```
etojeDenys, line 36:
```suggestion
  function getMother() {
  
  const mother = people.find(p => p.name === motherName);
  
    if (mother) {
      return (
            <a
              href={`#/people/${mother.slug}`}
              className="has-text-danger"
            >
              {mother.name}
            </a>
      );
    }

    return motherName || '-';
  }
```
you can simplify it

etojeDenys: and I suggest that you can use the Link component instead of the `a`

etojeDenys (changes requested):

SanyaBratashchuk (approved):

PR #668, opened 2023-12-23
Solution

IrynaK-a:
[DEMO LINK](https://IrynaK-a.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,12 +1,78 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+import { Link, useSearchParams } from 'react-router-dom';
+import cn from 'classnames';
+
+import { getSearchWith } from '../utils/searchHelper';
+import { FilterParams } from '../types/FilterParams';
+import { CENTURIES } from '../constants/centuries';
+
 export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const query = searchParams.get(FilterParams.Query) || '';
+
+  const handleChangeQuery = (e?: React.ChangeEvent<HTMLInputElement>) => {
+    const search = getSearchWith(
+      searchParams,
+      { [FilterParams.Query]: e?.target.value || null },
+    );
+
+    setSearchParams(search);
+  };
+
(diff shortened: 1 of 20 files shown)


src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,78 @@
...
+          )}
+          to={{
+            search: getSearchWith(searchParams, { [FilterParams.Sex]: null }),
+          }}
+        >
+          All
+        </Link>
+        <Link
+          className={cn(
+            { 'is-active': searchParams.get(FilterParams.Sex) === 'm' },
```
volodymyr-soltys97, line 58: Create a enum for this magic word 'm' and use it everywhere

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,78 @@
...
+          )}
+          to={{
+            search: getSearchWith(searchParams, { [FilterParams.Sex]: 'm' }),
+          }}
+        >
+          Male
+        </Link>
+        <Link
+          className={cn(
+            { 'is-active': searchParams.get(FilterParams.Sex) === 'f' },
```
volodymyr-soltys97, line 68: The same as above

src/components/PeopleFilters.tsx
```diff
@@ -16,77 +82,72 @@ export const PeopleFilters = () => {
...
+            value={query}
+            onChange={handleChangeQuery}
           />
 
           <span className="icon is-left">
-            <i className="fas fa-search" aria-hidden="true" />
+            <i
+              className="fas fa-search"
+              aria-hidden="true"
+              onClick={() => handleChangeQuery()}
```
volodymyr-soltys97, line 93:
```suggestion
              onClick={handleChangeQuery}
```

volodymyr-soltys97 (changes requested):
Great work 🔥 
Let's improve your code

src/components/PeopleFilters.tsx
IrynaK-a: As for me, it's better to leave this as it is, because in another way I need to add an additional event type and check for handleChangeQuery func

etojeDenys (approved):
lgtm

PR #670, opened 2023-12-26
Develop

MrHunterBoi:
[DEMO](https://mrhunterboi.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,12 +1,84 @@
+import { useSearchParams } from 'react-router-dom';
+import cn from 'classnames';
+
+import { Filter } from '../types/Filter';
+
 export const PeopleFilters = () => {
+  const [filterParams, setFilterParams] = useSearchParams();
+
+  const handleFilterChanging = (
+    field: Filter,
+    value: string,
+  ) => {
+    setFilterParams(currentParams => {
+      if (!value) {
+        currentParams.delete(field);
+
+        return currentParams;
+      }
+
+      currentParams.set(field, value);
+
+      return currentParams;
+    });
+  };
+
+  const handleCenturiesApplying = (value: string) => {
+    const centuries = filterParams.getAll('century');
+
+    if (centuries.includes(value)) {
+      setFilterParams(currentParams => {
+        currentParams.delete('century', value);
+
+        return currentParams;
+      });
+    } else {
+      setFilterParams(currentParams => {
+        currentParams.append('century', value);
+
+        return currentParams;
+      });
+    }
+  };
+
+  const handleResetCenturies = () => {
+    setFilterParams(currentParams => {
+      currentParams.delete('century');
+
+      return currentParams;
+    });
+  };
+
+  const handleResetAll = () => {
+    setFilterParams({});
+  };
+
   return (
     <nav className="panel">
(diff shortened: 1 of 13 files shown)


src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,91 @@
...
+          role="presentation"
+          className={cn({ 'is-active': !filterParams.get('sex') })}
+          onClick={() => handleFilterChanging(Filter.Sex, '')}
+        >
+          All
+        </a>
+        <a
+          role="presentation"
+          className={cn({ 'is-active': filterParams.get('sex') === 'm' })}
+          onClick={() => handleFilterChanging(Filter.Sex, 'm')}
```
roman-mirzoian, line 78: Avoid using "magic values"

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,91 @@
...
+        >
+          Male
+        </a>
+        <a
+          role="presentation"
+          className={cn({ 'is-active': filterParams.get('sex') === 'f' })}
+          onClick={() => handleFilterChanging(Filter.Sex, 'f')}
```
roman-mirzoian, line 85: Avoid using "magic values"

src/components/PeopleFilters.tsx
```diff
@@ -29,40 +113,60 @@ export const PeopleFilters = () => {
...
             <a
               data-cy="century"
-              className="button mr-1"
-              href="#/people?centuries=16"
+              role="presentation"
+              className={cn(
+                'button mr-1',
+                { 'is-info': filterParams.getAll('century').includes('16') },
+              )}
+              onClick={() => handleCenturiesApplying('16')}
```
roman-mirzoian, line 121: Buttons differ only in value (16, 17 etc.), so you can create an array with the necessary values and walk through it with `map` in order not to duplicate the code

src/components/PeopleFilters.tsx
```diff
@@ -29,40 +113,60 @@ export const PeopleFilters = () => {
           <div className="level-left">
             <a
```
roman-mirzoian, line 114: Errors from the linter must be fixed
image

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,122 @@
...
 
-              <p data-cy="noPeopleMessage">
-                There are no people on the server
-              </p>
+              {errorShowing && (
+                <p data-cy="peopleLoadingError">Something went wrong</p>
+              )}
 
-              <p>There are no people matching the current search criteria</p>
+              {/* <p>There are no people matching the current search criteria</p> */}
```
roman-mirzoian, line 107: This should be rendered if there are no search results, why is it commented out?

roman-mirzoian (changes requested):

src/components/PeopleFilters.tsx
MrHunterBoi: Oh, that's smart actually, haven't even thought about that

src/components/PeoplePage.tsx
MrHunterBoi: Oops

roman-mirzoian (approved):
Well done 👍

PR #672, opened 2023-12-29
finish task

yefimov77:
[DEMO LINK](https://yefimov77.github.io/react_people-table-advanced/)


src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,87 @@
...
+
+    if (!century) {
+      return getSearchWith(searchParams,
+        { [FilterParams.Century]: null });
+    }
+
+    if (century && centuriesParams.includes(century)) {
+      return getSearchWith(
+        searchParams,
+        { [FilterParams.Century]: centuriesParams.filter(c => c !== century) },
```
roman-mirzoian, line 34: Do not abbreviate names to unreadable ones

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,125 @@
...
+                  search: getSearchWith(searchParams, {
+                    [SORT]: SortType.Born,
+                    [SORT_ORDER]: getSortOrder(searchParams),
+                  }),
+                }}
+              >
+                <span className="icon">
+                  <i
+                    className={cn('fas',
+                      { [getArrowClass(searchParams, SortType.Born)]: true })}
```
roman-mirzoian, line 85: If there is no specific condition for the class, it can be written as the second argument, it is not necessary to specify the `true` condition, fix it in the entire code

src/components/PersonRow/Person.tsx
```diff
@@ -0,0 +1,69 @@
...
+    born,
+    died,
+    slug,
+    fatherName,
+    motherName,
+    mother,
+    father,
+  } = person;
+
+  const isFemale = sex === 'f';
```
roman-mirzoian, line 22: Avoid using "magic values"

src/helpers.ts
```diff
@@ -0,0 +1,112 @@
...
+import { Sex } from './types/Sextype';
+import { SortOrder } from './types/SortOrder';
+import { SortType } from './types/SortType';
+
+const NOT_SET_VALUE = '-';
+
+export const getPeopleWithParents = (people: Person[]) => {
+  return people.map(person => {
+    const mother = people.find(mom => person.motherName === mom.name);
+    const father = people.find(dad => person.fatherName === dad.name);
```
roman-mirzoian, line 12: At the time of the search, you don't yet know if it will be `dad` or `mom`
```suggestion
    const mother = people.find(({ name })=> person.motherName === name);
    const father = people.find(({ name }) => person.fatherName === name);
```

src/helpers.ts
```diff
@@ -0,0 +1,112 @@
...
+    );
+  }
+
+  if (sexFilter) {
+    peopleToRender = peopleToRender.filter(person => {
+      switch (sexFilter) {
+        case Sex.Male:
+          return person.sex === 'm';
+        case Sex.Female:
+          return person.sex === 'f';
```
roman-mirzoian, line 96: Magic values

roman-mirzoian (changes requested):
Also, in case of an unsuccessful search, a message about this should be displayed:
image

src/helpers.ts
```diff
@@ -0,0 +1,115 @@
+import { Person } from './types';
+import { FilterParams } from './types/FilterParams';
+import { Sex } from './types/Sextype';
+import { SortOrder } from './types/SortOrder';
+import { SortType } from './types/SortType';
+
+const NOT_SET_VALUE = '-';
+
+export const male = 'm';
+export const female = 'f';
```
roman-mirzoian, line 10: You already have an enum (Sex) with the same values, don't duplicate it

roman-mirzoian (approved):
Well done, but pay attention to the comment 👍

PR #674, opened 2023-12-29
add task solution

ViktorAfk:
[DEMO LINK](https://ViktorAfk.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,21 +1,70 @@
-export const PeopleFilters = () => {
+import cn from 'classnames';
+import { Options, PersonSex } from '../types/enum';
+import { SearchLink } from './SearchLink';
+
+const centuriesForRender:string[] = ['16', '17', '18', '19', '20'];
+
+interface Props {
+  query: string;
+  handleQueryChange:(value: React.ChangeEvent<HTMLInputElement>) => void;
+  centuries: string[];
+  sex: string;
+}
+
+export const PeopleFilters: React.FC<Props> = (props) => {
+  const {
+    query,
+    handleQueryChange,
+    centuries,
+    sex,
+  } = props;
+
+  const setCenturies = (century:string) => (
+    {
+      centuries: centuries
+        .includes(century.toString())
+        ? centuries
+          .filter(currentCentury => currentCentury !== century.toString())
+        : [...centuries, century],
+    }
+  );
+
+  const onReset = () => ({ centuries: null, query: null, sex: null });
+
   return (
     <nav className="panel">
       <p className="panel-heading">Filters</p>
 
       <p className="panel-tabs" data-cy="SexFilter">
-        <a className="is-active" href="#/people">All</a>
-        <a className="" href="#/people?sex=m">Male</a>
-        <a className="" href="#/people?sex=f">Female</a>
+        <SearchLink
+          params={{ sex: null }}
+          className={cn({ 'is-active': !sex })}
+        >
+          {Options.ALL}
+        </SearchLink>
+        <SearchLink
+          params={{ sex: PersonSex.MALE }}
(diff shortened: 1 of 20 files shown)


ruslanagnatenko (approved):

src/components/PeopleFilters.tsx
```diff
@@ -1,21 +1,70 @@
...
+import { Options, PersonSex } from '../types/enum';
+import { SearchLink } from './SearchLink';
+
+const centuriesForRender:string[] = ['16', '17', '18', '19', '20'];
+
+interface Props {
+  query: string;
+  handleQueryChange:(value: React.ChangeEvent<HTMLInputElement>) => void;
+  centuries: string[];
+  sex: string;
```
BudnikOleksii, line 11:
```suggestion
  sex: PersonSex.MALE;
```
Use your enum

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,78 @@
...
+                : (isError && <PeopleLoadingError />) || (people.length > 0
+                  ? (
+                    <PeopleTable
+                      people={filteredPeople}
+                      sort={sort}
+                      order={order}
+                      setSearchWith={setSearchWith}
+                    />
+                  )
+                  : <NoPeopleMessage />)}
```
BudnikOleksii, line 75: Avoid nested conditions. Simplify them.
When isLoading show loader, when isError - error, etc. Without ternary operators

src/index.tsx
```diff
@@ -1,14 +1,31 @@
...
+        <Route path="/home" element={<Navigate to="/" replace />} />
+        <Route path="/" element={<App />}>
+          <Route index element={<h1 className="title">Home Page</h1>} />
+          <Route path="people/personSlug?">
+            <Route index element={<PeoplePage />} />
+            <Route path=":personSlug" element={<PeoplePage />} />
+          </Route>
+          <Route path="*" element={<NotFoundPage />} />
+        </Route>
+      </Routes>
```
BudnikOleksii, line 29: Create a separate component for your routes

src/utils/helpers.ts
```diff
@@ -0,0 +1,86 @@
+import { Person, Title } from '../types';
+import { PersonSex } from '../types/enum';
+
+const getFieldFromTitle = (column:Title, columnCode:string) => {
+  switch (columnCode) {
+    case 'name':
```
BudnikOleksii, line 6: Create an enum for magic values

src/utils/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+};
+
+export const getFilteredPeople = (
+  people: Person[],
+  query: string,
+  sortBySex: string,
+  centuries: string[],
+  sort: string,
+  order: string,
+): Person[] => {
```
BudnikOleksii, line 30: The function looks too complicated, split logic, for example, one function for filter and the second for sorting

BudnikOleksii (changes requested):
Good work 👍 
But could be even better, check the suggesitions

src/components/PeopleFilters.tsx
ViktorAfk: That I can't fix. It comes from the searchParam type that was set up in the project before

BudnikOleksii: > That I can't fix. It comes from the searchParam type that was set up in the project before

Got it 👍 But why you can't fix it if you created an enum?)

BudnikOleksii (approved):
Great👍

src/components/PeopleFilters.tsx
ViktorAfk: I don't know. I have such a mistake in props 
Знімок екрана 2023-12-30 о 09 54 12

PR #694, opened 2024-01-05
Solution

Barosz30:
https://barosz30.github.io/react_people-table-advanced/


mbruhler (changes requested):
I don't like the fact that when I input something to query field, and then I want to go back  it changes letter by letter. It should in my opinion work as one go back in browser history

mpiskadlo (changes requested):
Clicking on an individual person clears all filters, it's not the case in [MATE DEMO](https://mate-academy.github.io/react_people-table-advanced/#/)

mpiskadlo (approved):
All good!

PR #705, opened 2024-01-15
Solution

MarcinLigmann:
[DEMO LINK](https://MarcinLigmann.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -91,7 +91,7 @@ export const PeopleFilters = () => {
                 key={el}
                 data-cy="century"
                 className={classNames('button mr-1', {
-                  'is-info': centuries?.includes(String(el)),
+                  'is-info': centuries.includes(String(el)),
                 })}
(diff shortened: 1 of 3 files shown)


bertsu02 (approved):

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,69 @@
+import { useSearchParams } from 'react-router-dom';
+import classNames from 'classnames';
+import { useState } from 'react';
+import { SearchLink } from './SearchLink';
+
 export const PeopleFilters = () => {
+  const CENTURIES_VALUE: number[] = [16, 17, 18, 19, 20];
```
mpiskadlo, line 7: I would keep it outside of the component, typically in some kind of config file.

BTW you don't need that `number[]` here. TS would figure it out by itself ;)

mpiskadlo (approved):
Well done!

I just had one small suggestion, but is not worth requesting changes ;)

PR #706, opened 2024-01-15
Solution

denedweb:
[DEMO LINK](https://denedweb.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PersonLink.tsx
+++ b/src/components/PersonLink.tsx
@@ -1,0 +1,22 @@
+import { Link } from 'react-router-dom';
+import classNames from 'classnames';
+import { Person } from '../types';
+
+type Props = {
+  person: Person,
+};
+
+export const PersonLink: React.FC<Props> = ({
+  person,
+}) => {
+  return (
+    <Link
+      to={`../${person.slug}`}
+      className={classNames(
+        { 'has-text-danger': person.sex === 'f' },
+      )}
+    >
+      {person.name}
+    </Link>
+  );
+};
--- a/src/utils/filterPeople.tsx
+++ b/src/utils/filterPeople.tsx
@@ -1,0 +1,82 @@
+import { Person } from '../types';
+import { Sort } from '../types/Sort';
+
+type PropsParams = {
+  query: string | null,
+  sort: string | null,
+  order: string | null,
+  sex: string | null,
+  centuries: string[],
+};
+
+export const getPreparedPeople = (
+  people: Person[], {
+    query,
+    sort,
+    order,
+    sex,
+    centuries,
+  }: PropsParams,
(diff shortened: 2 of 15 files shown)


src/utils/filterPeople.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  if (sort) {
+    preparedPeople = preparedPeople.sort((person1, person2) => {
+      switch (sort) {
+        case Sort.Name:
+          return person1.name.localeCompare(person2.name);
+
+        case Sort.Sex:
+          return person1.sex.localeCompare(person2.sex);
+
```
etojeDenys, line 41:
```suggestion
      switch (sort) {
        case Sort.Name:
        case Sort.Sex:
          return person1[sort].localeCompare(person2[sort]);

```
we can group it

src/utils/filterPeople.tsx
```diff
@@ -0,0 +1,83 @@
...
+        case Sort.Born:
+          return person1.born - person2.born;
+
+        case Sort.Died:
+          return person1.died - person2.died;
```
etojeDenys, line 46: same here

etojeDenys (changes requested):
1. this message looks redundant here
image

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,22 @@
...
+};
+
+export const PersonLink: React.FC<Props> = ({
+  person,
+}) => {
+  return (
+    <Link
+      to={`../${person.slug}`}
+      className={classNames(
+        { 'has-text-danger': person.sex === 'f' },
```
volodymyr-soltys97, line 16: You can use a enum here instead of 'f'

src/utils/filterPeople.tsx
```diff
@@ -0,0 +1,79 @@
...
+    });
+  }
+
+  if (sex) {
+    switch (sex) {
+      case Sex.All:
+        return preparedPeople;
+
+      case Sex.Male:
+        return preparedPeople.filter(person => person.sex === 'm');
```
volodymyr-soltys97, line 64: And here, use a enum for 'm'

src/utils/filterPeople.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+      case Sex.Female:
+        return preparedPeople.filter(person => person.sex === 'f');
```
volodymyr-soltys97, line 67: The same as above, need to fix it everywhere

volodymyr-soltys97 (changes requested):
Almost done!

maxim2310 (approved):
Good work!!!

PR #709, opened 2024-01-16
Develop

jacekkosiba:
[DEMO LINK](https://jacekkosiba.github.io/react_people-table-advanced/)


szymanski-marcin (approved):
Looks awesome. Great work!

danon321 (changes requested):
image
There u have to statuses of data: loading and 'no found' in the same time

image
Like in example in this case table should disappear and only stay "There are no people matching the current search criteria"

jacekkosiba:
thanks for pointing this out!
should be all right now :)

danon321 (approved):

PR #710, opened 2024-01-16
add task solution

MariaSnegireva:
[DEMO LINK](https://MariaSnegireva.github.io/react_people-table-advanced/)


src/components/HomeRedirect.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { useNavigate } from 'react-router-dom';
+
+export const HomeRedirect = () => {
+  const navigate = useNavigate();
+
+  useEffect(() => {
+    navigate('/', { replace: true });
+  }, [navigate]);
+
+  return (<></>);
```
witflash, line 11: For such case you can just `return null`. There is no sense to return a Fragment

witflash (approved):
Great work! I like it.
Well done 👍🏻

PR #712, opened 2024-01-17
solution

Timi2424:
[DEMO LINK](https://Timi2424.github.io/react_people-table-advanced/)


marcinzrodlowski (approved):
Based on my previous experiences, even when all tests are passing, there is always a chance of something being wrong. Personally, I think the app looks great, tests are passing, so good job! 💪

Timi2424:
> Based on my previous experiences, even when all tests are passing, there is always a chance of something being wrong. Personally, I think the app looks great, tests are passing, so good job! 💪

there were no tests implemented yet for filters

PR #713, opened 2024-01-17
Solution

marcinzrodlowski:
[DEMO LINK](https://marcinzrodlowski.github.io/react_people-table-advanced/)


adrian200002 (approved):
Works good. The only small issue that i noticed is the color of the reset all filters button. It's a bit different in original example.

src/components/PeoplePage.tsx
```diff
@@ -1,31 +1,138 @@
...
+
+export const PeoplePage: FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [hasError, setHasError] = useState(false);
+
+  const [searchParams] = useSearchParams();
+  const query = searchParams.get(QueryTypes.Query) || '';
+  const sex = searchParams.get(QueryTypes.Sex) || '';
+  const centuries = searchParams.getAll(QueryTypes.Centuries) || [];
```
mpiskadlo, line 21: If there is no such a params, centuries would be an empty array anyways

Don't worry. It's not worth requesting changes 😅 

```suggestion
  const centuries = searchParams.getAll(QueryTypes.Centuries);
```

mpiskadlo (approved):
Good job!

PR #717, opened 2024-01-19
Solution

SashaViun:
[DEMO LINK](https://SashaViun.github.io/react_people-table-advanced/)


etojeDenys (changes requested):
1. why do we have two 17 buttons and one of them always selected
image

DarkMistyRoom (changes requested):
Good job! Only one comment:
When I apply sorting or/and search params and then select some person - all params reset. Please keep them after selecting

PR #720, opened 2024-01-21
demo

OPTIMISTIXX:
[DEMO LINK](https://OPTIMISTIXX.github.io/react_people-table-advanced/)


VitaliyBondarenko1982 (changes requested):
Demo link not works (leads to 404 page)
Try to solve cycle dependencies issue. For it make for context separate file and import it in all places where it need. Now problem that in App you imports PeoplePage (whick imports PeopleTable), but in PeopleTable exists import from App

Screenshot from 2024-01-21 16-44-18

PR #736, opened 2024-02-07
add solution

HokageHEL:
[DEMO LINK](https://HokageHEL.github.io/react_people-table-advanced/)


src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,110 @@
...
+      filteredPeople?.sort((a, b) => a.sex.localeCompare(b.sex));
+    }
+
+    if (sortType === 'born') {
+      filteredPeople?.sort((a, b) => a.born - b.born);
+    }
+
+    if (sortType === 'died') {
+      filteredPeople?.sort((a, b) => a.died - b.died);
+    }
```
etojeDenys, line 62: better to use switch case here, you can group some cases and simplify the logic

etojeDenys (changes requested):
https://github.com/mate-academy/react_people-table-advanced/assets/70220667/f263e8d0-b262-46c3-be29-cb14746c935a

looks like something is wrong with the icons, when I click on the name, you need to change only "name" icon


2. looks different
image
image

src/components/PeopleFilters.tsx
```diff
@@ -1,12 +1,71 @@
...
+            'is-active': !sex,
+          })}
+          to={{
+            search: getSearchWith(searchParams, { sex: null }),
+          }}
+        >
+          All
+        </Link>
+
+        <Link
```
TarasHoliuk, line 45: not a mistake, just FYI - there was a `SearchLink` component to do the same :)

TarasHoliuk (approved):
Well done :heavy_check_mark: 

Clear code :+1:

PR #767, opened 2024-02-27
Develop

zvir91:
[DEMO LINK](https://zvir91.github.io/react_people-table-advanced/)


src/Pages/PeoplePage.tsx
```diff
@@ -0,0 +1,219 @@
...
+  const [searchParams] = useSearchParams();
+
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+  const sex = searchParams.get('sex') || '';
+  const sort = searchParams.get('sort') || '';
+  const order = searchParams.get('order') || '';
+
+  const filteredPeople =
+    people && filterPeople(people, query, centuries, sex, sort, order);
```
witflash, line 33: You don't need to check the `people` because your `people` variable is always an array (even when people don't exist at all, this variable will be an empty array).

src/Pages/PeoplePage.tsx
```diff
@@ -0,0 +1,219 @@
...
+              <PeopleFilters />
+            </div>
+          )}
+
+          <div className="column">
+            <div className="box table-container">
+              {loading ? (
+                <Loader />
+              ) : (
+                isPeopleList && (
```
witflash, line 81: I like the next syntax:
```
{loading && <Loader />}
{!loading && isPeopleList && (...)}
```
It's more readable for me.
You can try it in your future work if you also like such an approach.

src/components/PeopleFilters.tsx
```diff
@@ -32,64 +72,47 @@ export const PeopleFilters = () => {
...
-
-            <a
-              data-cy="century"
-              className="button mr-1"
-              href="#/people?centuries=20"
-            >
-              20
-            </a>
-          </div>
+          {['16', '17', '18', '19', '20'].map(num => (
```
witflash, line 75: It's better to move this array to the separate constant outside of the React Component (in this file, but just above where you declare `PeopleFilters`).
It will prevent unnecessary render (because now this array will be created from scratch on any changes in your component). 
It's just about performance improvements.

src/services/filter.ts
```diff
@@ -0,0 +1,87 @@
...
+    filteredPeople = filteredPeople.filter(person => person.sex === 'm');
+  }
+
+  switch (true) {
+    case sort === SortType.Born && order === 'desc':
+      filteredPeople = filteredPeople.sort(
+        (personA, personB) => personB.born - personA.born,
+      );
+
+      break;
```
witflash, line 38: You can avoid using `break` for every `case` if you return something:
```
case .....:
  return filteredpeople = ...
```

witflash (approved):
Great work!
I've just left several comments for your future improvements.

PR #772, opened 2024-03-01
solution

mariuszgit:
https://mariuszgit.github.io/react_people-table-advanced/


src/App.tsx
```diff
@@ -10,9 +13,16 @@ export const App = () => {
...
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
+            <Route path="/" element={<HomePage />} />
+
+            <Route path="/home" element={<Navigate to="/" replace />} />
+
+            <Route path="/people/" element={<PeoplePage />} />
```
yevhenii-pyl, line 21: As a general rule of thumb implementation: 
```
              <Route path="people">
                <Route path=":slug?" element={<PeoplePage />} />
              </Route>
```

This way you implicitly state that 'hey, there is a root page and it may include slug.

src/components/PeopleFilters.tsx
```diff
@@ -22,6 +49,18 @@ export const PeopleFilters = () => {
             type="search"
             className="input"
             placeholder="Search"
+            value={activeQuery}
+            onChange={e => {
```
yevhenii-pyl, line 53: It definitely wants to be a separate handler function

src/components/Table.tsx
```diff
@@ -0,0 +1,281 @@
...
+          </th>
+
+          <th>Mother</th>
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {filteredPeople.map(
+          ({
```
yevhenii-pyl, line 226: 👍

src/components/Table.tsx
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    if (activeOrder === 'desc') {
+      tempPeople.reverse();
+    }
+
+    return tempPeople;
+  }, [computedPeople, activeSort, activeOrder]);
+
+  const filteredPeople = useMemo(() => {
```
yevhenii-pyl, line 72: I'm not sure that any of these useMemo will lead to render optimisation. 
This hook uses shallow comparison for dependency array and here each dependancy array contains an array as a dependency. Which means, each time this component receives people prop => these 'memoized' values will be recalculated. 
In these cases caching looks like a more expensive calculation than the actual calculation :)

src/hooks/usePeople.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Person } from '../types';
+
+const usePeople = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+
+  useEffect(() => {
+    const getData = async () => {
+      setIsLoading(true);
```
yevhenii-pyl, line 12: Setting the loading state can also be placed inside `try` block

src/components/PeopleFilters.tsx
```diff
@@ -1,18 +1,45 @@
+import { useSearchParams } from 'react-router-dom';
+import cn from 'classnames';
+import { SearchLink } from './SearchLink';
+
 export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const centuries = [16, 17, 18, 19, 20];
```
yevhenii-pyl, line 7: Should be declared outside the component, let's assume we are taking this value from environmental variables and declare it in uppercase.

yevhenii-pyl (changes requested):
Solid work here! 🥇 
Let's adjust a few minor things and we are good to go

src/components/Table.tsx
mariuszgit: I removed useMemo from filteredPeople becouse it doesnt change much, but leaved on sorting and computed people becouse I dont want to compute them every time I type something - especialy whe people array is very big

yevhenii-pyl (approved):
Nice, reviewing this was simply a pleasure 🥇

PR #789, opened 2024-03-18
add task solution

denlubn:
[DEMO LINK](https://denlubn.github.io/react_people-table-advanced/)


src/utils/getFilteredPeople.ts
```diff
@@ -0,0 +1,92 @@
...
+      break;
+    }
+
+    case SortType.Sex: {
+      preparedPeople = preparedPeople.sort((person1, person2) =>
+        person1.sex.localeCompare(person2.sex),
+      );
+
+      break;
+    }
```
etojeDenys, line 39:
```suggestion
    case SortType.Name:
    case SortType.Sex:
      preparedPeople = preparedPeople.sort((person1, person2) =>
        person1[sort].localeCompare(person2[sort]),
      );

      break;
```
we can group some cases

src/utils/getFilteredPeople.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    case SortType.Died: {
+      preparedPeople = preparedPeople.sort(
+        (person1, person2) => person1.died - person2.died,
+      );
+
+      break;
+    }
```
etojeDenys, line 55: same here

etojeDenys (changes requested):

PR #799, opened 2024-03-26
Solution

oshapkun:
[DEMO LINK](https://oshapkun.github.io/react_people-table-advanced/)


src/components/PeopleFilters.tsx
```diff
@@ -1,18 +1,62 @@
...
+import { SexStatus } from '../types/SexStatus';
+
 export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const sex = searchParams.get('sex') || '';
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+
+  const setSearchWith = (params: any) => {
```
GUSILLUS, line 13: Try to avoid using any here

GUSILLUS (changes requested):
GJ! Try to fix test and avoid using any type

sTorba24 (changes requested):
Good work!

Now if i click other link, it'll reset query params, but it shouldn't, you need to save all settings , only name change(i think it is the root why you didn't pass tests)

If you have any issues with it - feel free to text your fe_chat

image

maxim2310 (changes requested):
almost done try to pass the remaining tests

nazarmatsevych (approved):
Good work!

PR #801, opened 2024-03-28
update template, add solution

dimakrugly:
[DEMO LINK](https://dimakrugly.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,23 +1,67 @@
+import cn from 'classnames';
+import { useSearchParams } from 'react-router-dom';
+import { SearchLink } from './SearchLink';
+import { ARRAY_OF_CENTURIES } from '../utils/constants';
+
 export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const handleSetName = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const params = new URLSearchParams(searchParams);
+
+    if (event.target.value) {
+      params.set('query', event.target.value);
+    } else {
+      params.delete('query');
+    }
+
+    setSearchParams(params);
+  };
+
+  const century = searchParams.getAll('centuries') || [];
+  const query = searchParams.get('query') || '';
+  const sex = searchParams.get('sex') || '';
+
+  const toggleCentury = (centuryProp: string) => {
+    const allCenturies = century.includes(centuryProp)
+      ? century.filter(c => c !== centuryProp)
+      : [...century, centuryProp];
+
+    return allCenturies;
+  };
+
   return (
     <nav className="panel">
       <p className="panel-heading">Filters</p>
 
       <p className="panel-tabs" data-cy="SexFilter">
-        <a className="is-active" href="#/people">
+        <SearchLink
+          params={{ sex: null }}
+          className={cn({ 'is-active': !sex })}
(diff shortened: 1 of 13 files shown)


src/components/PeopleTable.tsx
```diff
@@ -1,5 +1,127 @@
...
+    }
+
+    if (sortParam === param && orderParam === 'desc') {
+      return 'fa-sort-down';
+    }
+
+    return '';
+  };
+
+  const filterPeople = useCallback(() => {
```
TarasHoliuk, line 53:
```suggestion
  const filteredPeople = useMemo(() => {
```
And instead of `setFilteredPeople(filtered);` return `filtered`;

filteredPeople state is redundant - delete it

src/components/PersonLInk.tsx
```diff
@@ -0,0 +1,79 @@
...
+  filteredPeople: Person[];
+};
+
+export const PersonLink: React.FC<Props> = ({ person, filteredPeople }) => {
+  const { born, died, fatherName, motherName, name, sex, slug } = person;
+
+  const { slugParam } = useParams();
+  const [searchParams] = useSearchParams();
+
+  const hasMother = useMemo(
```
TarasHoliuk, line 18:
```suggestion
  const maybeMother = useMemo(
```
`has` sounds like it's boolean value

src/components/PersonLInk.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </td>
+      <td>{sex}</td>
+      <td>{born}</td>
+      <td>{died}</td>
+      <td>
+        {hasMother ? (
+          <Link
+            className={cn({ 'has-text-danger': hasMother.sex === 'f' })}
+            to={{
+              pathname: `/people/${hasMother.slug}`,
```
TarasHoliuk, line 53: note that one test related to this link failed. Try to fix it

TarasHoliuk (changes requested):
Great job
Added a few small comments

src/components/PersonLInk.tsx
dimakrugly: 1) 
       #/people/:correct-slug page
         should allow to select another person:
     AssertionError: Timed out retrying after 4000ms: Expected to find element: `[data-cy="person"]`, but never found it.
      at Context.eval (webpack://react_people-table-advanced/./cypress/support/commands.ts:39:0)
      
      
      
      
      I have no idea how to fix it. It says that I'm missing the data-cy attribute, but I'm not. All the tests passed well in the previous version of the boilerplate.

src/components/PeopleFilters.tsx
```diff
@@ -1,23 +1,67 @@
...
+          params={{ sex: null }}
+          className={cn({ 'is-active': !sex })}
+        >
           All
-        </a>
-        <a className="" href="#/people?sex=m">
+        </SearchLink>
+
+        <SearchLink
+          className={cn({ 'is-active': sex === 'm' })}
```
volodymyr-soltys97, line 46: Create a enum for 'm' and 'f' and use it everywhere

volodymyr-soltys97 (changes requested):
GJ!
Let's make your code better
Try to fix this problem with tests, if you need help feel free in the fe_chat
image

volodymyr-soltys97 (approved):
GJ!

PR #813, opened 2024-04-04
react_people-table-advanced

Hovera-Liubomyr:
https://Hovera-Liubomyr.github.io/react_people-table-advanced/

### DIFF
--- a/src/utils/Filter.ts
+++ b/src/utils/Filter.ts
@@ -1,0 +1,34 @@
+import { Person } from '../types';
+
+export const Filter = (people: Person[], searchParams: URLSearchParams) => {
+  const searchSex = searchParams.get('sex') || '';
+  const searchCenturies = searchParams.getAll('century') || [];
+  const searchQuery = searchParams.get('query') || '';
+
+  if (!searchSex && searchCenturies.length === 0 && !searchQuery) {
+    return people;
+  }
+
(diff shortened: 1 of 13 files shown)


maxim2310 (changes requested):
Good work, but after you select some filter and select person your previous filters are refreshed

src/utils/Filter.ts
```diff
@@ -0,0 +1,34 @@
+import { Person } from '../types';
+
+export const Filter = (people: Person[], searchParams: URLSearchParams) => {
```
etojeDenys, line 3:
```suggestion
export const filterPeople = (people: Person[], searchParams: URLSearchParams) => {
```
name functions in camelCase

src/utils/Sort.ts
```diff
@@ -0,0 +1,33 @@
+import { Person } from '../types';
+
+export const Sort = (people: Person[], searchParams: URLSearchParams) => {
```
etojeDenys, line 3:
```suggestion
export const sortPeople = (people: Person[], searchParams: URLSearchParams) => {
```

etojeDenys (changes requested):
image
don't show the message in this case

2. looks different from the working example
image

etojeDenys (approved):
good job

PR #818, opened 2024-04-07
solution

Diliradon:
https://diliradon.github.io/react_people-table-advanced/

### DIFF
--- a/src/components/PeoplePage.tsx
+++ b/src/components/PeoplePage.tsx
@@ -1,29 +1,53 @@
+import { useEffect, useState } from 'react';
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { getPeople } from '../api';
+import { Person } from '../types';
 
 export const PeoplePage = () => {
+  const [isLoader, setIsLoader] = useState(true);
+  const [isError, setIsError] = useState(false);
+  const [people, setPeople] = useState<Person[]>([]);
+
+  useEffect(() => {
+    getPeople()
+      .then(setPeople)
+      .catch(() => setIsError(true))
+      .finally(() => setIsLoader(false));
+  }, []);
+
   return (
     <>
       <h1 className="title">People Page</h1>
 
       <div className="block">
         <div className="columns is-desktop is-flex-direction-row-reverse">
-          <div className="column is-7-tablet is-narrow-desktop">
-            <PeopleFilters />
-          </div>
+          {!isLoader && (
+            <div className="column is-7-tablet is-narrow-desktop">
+              <PeopleFilters />
+            </div>
+          )}
 
           <div className="column">
             <div className="box table-container">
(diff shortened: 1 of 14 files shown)


src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,53 @@
...
+              {!people.length && !isLoader && (
+                <p data-cy="noPeopleMessage">
+                  There are no people on the server
+                </p>
+              )}
 
-              <p>There are no people matching the current search criteria</p>
+              {!people.length && (
+                <p>Th ere are no people matching the current search criteria</p>
+              )}
```
volodymyr-soltys97, line 48: This text is shown when people download, it does not need to be shown during the download process
```suggestion
              {!people.length && (
                <p>There are no people matching the current search criteria</p>
              )}
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
Filters are now reset after we click on a person, this needs to be fixed
image

IvanFesenko (changes requested):
- On first load should be only loading without error message
2024-04-09_23h33_23
- click on person should not reset filters
https://github.com/mate-academy/react_people-table-advanced/assets/61904995/8c31338a-048a-41be-9db0-9bfa836cd3b4

l4st1m0za (changes requested):
You still have not change the bug, when on the click on the person, all the filters are erased

Diliradon:
I forgot to wirite npm run deploy

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,130 @@
...
+
+      return 'fas fa-sort-down';
+    }
+
+    return 'fas fa-sort';
+  };
+
+  const getCentury = (year: number) => Math.floor(year / 100) + 1;
+
+  const preparePeopleList = [...people]
```
TarasHoliuk, line 47:
```suggestion
  const preparedPeopleList = [...people]
```
it's array - already prepared, not a function

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,130 @@
...
+    .sort((p1, p2) => {
+      let res = 0;
+
+      switch (sort) {
+        case 'name':
+          res = p1.name.localeCompare(p2.name);
+          break;
+        case 'sex':
+          res = p1.sex.localeCompare(p2.sex);
+          break;
```
TarasHoliuk, line 57:
```suggestion
        case 'name':
        case 'sex':
          res = p1.[sort]localeCompare(p2[sort]);
          break;
```
you could join some cases

src/components/PeoplePage.tsx
TarasHoliuk, line 47: It made sense to filter first (it'd return copy) and then sort

TarasHoliuk (approved):
GJ 👍 

Approved ✔️ Added a few comments just FYI

PR #819, opened 2024-04-09
solution

Arkadiuszx:
[DEMO LINK](https://Arkadiuszx.github.io/react_people-table-advanced/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,10 +6,10 @@ import './App.scss';
 export const App = () => {
   return (
     <Routes>
-      <Route element={<Layout />}>
-        <Route path="/" element={<h1 className="title">Home Page</h1>} />
-        <Route path="/people" element={<PeoplePage />}>
-          <Route path=":personSlug" element={<PeoplePage />} />
+      <Route path="/" element={<Layout />}>
+        <Route index element={<h1 className="title">Home Page</h1>} />
+        <Route path="/people">
+          <Route path=":personSlug?" element={<PeoplePage />} />
         </Route>
         <Route path="/home" element={<Navigate to="/" replace />} />
         <Route path="*" element={<h1 className="title">Page not found</h1>} />


src/App.tsx
```diff
@@ -10,9 +11,25 @@ export const App = () => {
...
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
+            <Route path="/" element={<h1 className="title">Home Page</h1>} />
+            <Route
+              path="/people"
+              element={
+                <>
+                  <h1 className="title">People Page</h1>
```
yevhenii-pyl, line 20: I'd prefer to have the title inside the page, not as a separate part in routes.

src/App.tsx
```diff
@@ -2,6 +2,7 @@ import { PeoplePage } from './components/PeoplePage';
 import { Navbar } from './components/Navbar';
```
yevhenii-pyl, line 2: Would be better to create a separate 'Layout' or 'Base' component for such common ...well, layout. 
If you want to share some layout between all your page components, this should be done like this: 

```
export const Layout = () => {
  return (
    <div data-cy="app">
      <Navbar />

      <main className="section">
        <div className="container">
          <Outlet />
        </div>
      </main>
    </div>
  );
};
```

And used in routes like: 

```
      <Route path="/" element={<Layout />}>
        <Route path="people">
          <Route path=":personSlug?" element={<PeoplePage />} />
        </Route>
      </Route>
```

src/components/PeopleFilter/PeopleFilters.tsx
```diff
@@ -0,0 +1,118 @@
+import { useSearchParams } from 'react-router-dom';
```
yevhenii-pyl, line 1: Don't forget to place your imports accordingly:

```
third-party libraries, starting from react

common modules 

inner modules 
```

in this case: 

```
import React from 'react'
import cn from 'classnames';
...
import { SearchLink } from '../SearchLink';
```

src/components/PeoplePage.tsx
```diff
@@ -1,12 +1,41 @@
...
 export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState(false);
+  const [searchParams] = useSearchParams();
+
+  useEffect(() => {
+    const load = async () => {
+      try {
+        setLoading(true);
```
yevhenii-pyl, line 20: 👍

src/components/PeoplePage.tsx
```diff
@@ -15,15 +44,27 @@ export const PeoplePage = () => {
...
 
-              <p data-cy="peopleLoadingError">Something went wrong</p>
+              {error && (
+                <p data-cy="peopleLoadingError" className="has-text-danger">
+                  Something went wrong
+                </p>
+              )}
 
-              <p data-cy="noPeopleMessage">There are no people on the server</p>
+              {people && !people?.length && !loading && (
```
yevhenii-pyl, line 55: `!people?.length && ! loading ` should be enough for this check

yevhenii-pyl (changes requested):
Great job!  🚀 
Your code becomes better and better each time. 

Let's adjust your router to best practices and we are good to go.

src/components/Navbar.tsx
```diff
@@ -8,17 +12,19 @@ export const Navbar = () => {
     >
       <div className="container">
         <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
+          <NavLink
+            className={`navbar-item ${pathname === '/' ? 'has-background-grey-lighter' : ''}`}
```
BogdanMaliuta, line 16: for conditional styles use library classNames

src/components/PeopleTable/TableHeader.tsx
```diff
@@ -0,0 +1,49 @@
...
+    );
+
+  const linkParam = (param: string) => {
+    return {
+      sort: param,
+      order: sort === param ? (order === 'desc' ? null : 'desc') : null,
+      ...(sort === param && order === 'desc'
+        ? { sort: null, order: null }
+        : {}),
+    };
```
BogdanMaliuta, line 27: its not good idea to use ternary operators in ternary operator, this 3 lines is very hard to read, try to make it in another way

BogdanMaliuta (approved):
Hello, good code, but in my opinion "ternary" solution needs to be changed, Also use classNames library if you needs cond styles

PR #826, opened 2024-04-11
solution

IRONH4MMER:
[DEMO LINK](https://IRONH4MMER.github.io/react_people-table-advanced/)


src/components/People/PeopleFilters.tsx
```diff
@@ -0,0 +1,132 @@
...
+    setSearchWith({ query: event.target.value || null });
+  };
+
+  const clearInput = () => {
+    setSearchWith({ query: null });
+  };
+
+  const sexFilters = [
+    {
+      name: 'All',
```
polosanya, line 38: Would be great to use Enum instead of hardcoded values

src/components/People/PeopleTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+const sortPeople = (
+  people: Person[],
+  sort: string | null,
+  order: string | null,
+) => {
+  let sortedPeople = people;
+
+  switch (sort) {
+    case 'name':
```
polosanya, line 24: Same here. Use Enums

polosanya (changes requested):
Good job! Let's improve your code readability a bit)

PR #849, opened 2024-04-16
finished people table

julwer1k:
[DEMO](https://julwer1k.github.io/react_people-table-advanced/)


src/components/People/PeopleTable.tsx
```diff
@@ -0,0 +1,228 @@
...
+        case TitleTableHeaders.name:
+          sortedPeople = sortedPeople.sort((person1, person2) =>
+            person1.name.localeCompare(person2.name),
+          );
+          break;
+        case TitleTableHeaders.sex:
+          sortedPeople = sortedPeople.sort((person1, person2) =>
+            person1.sex.localeCompare(person2.sex),
+          );
+          break;
```
etojeDenys, line 103:
```suggestion
        case TitleTableHeaders.name:
        case TitleTableHeaders.sex:
          sortedPeople = sortedPeople.sort((person1, person2) =>
            person1[sort].localeCompare(person2[sort]),
          );
          break;
```
we can group some cases

src/components/People/PeopleTable.tsx
```diff
@@ -0,0 +1,228 @@
...
+        case TitleTableHeaders.born:
+          sortedPeople = sortedPeople.sort(
+            (person1, person2) => person1.born - person2.born,
+          );
+          break;
+        case TitleTableHeaders.died:
+          sortedPeople = sortedPeople.sort(
+            (person1, person2) => person1.died - person2.died,
+          );
+          break;
```
etojeDenys, line 113: same here

etojeDenys (changes requested):
https://github.com/mate-academy/react_people-table-advanced/assets/70220667/e33d088e-f629-4129-bd1b-6f71ab8cd05f

do not reset the filters on person click


2. looks like filtering by query doesn't work, try to fix that
image

src/utils/getPreparedPeople.ts
```diff
@@ -0,0 +1,14 @@
+import { Person } from '../types';
+
+export const getPreparedPeople = (people: Person[]) => {
+  let preparedPeople = [...people];
+
+  preparedPeople = preparedPeople.map(person => {
```
TarasHoliuk, line 6:
```suggestion
  return preparedPeople.map(person => {
```
`map` returns copy of array, so you could do this

src/types/Person.ts
```diff
@@ -1,4 +1,5 @@
 export interface Person {
+  [key: string]: any;
```
TarasHoliuk, line 2: seems redundant

src/components/People/PeopleTable.tsx
```diff
@@ -0,0 +1,249 @@
...
+    if (sort && order) {
+      return { sort: null, order: null };
+    }
+
+    return { sort: null, order: null };
+  };
+
+  const preparedPeople = getPreparedPeople(people);
+
+  const getSortedPeople = () => {
```
TarasHoliuk, line 88: if it's quite a big reusable function it makes sense to move it out of components and accept all parameters you need as arguments. But in this case would be even better to use `useMemo`

src/components/People/PeopleTable.tsx
```diff
@@ -0,0 +1,249 @@
...
+  const order = searchParams.get('order') || null;
+
+  const sex = searchParams.get('sex') || null;
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+
+  const showTable = !isLoading && !isError && isMatchSearch;
+  const showSideBar = !isLoading && !isError;
+
+  const setSearchWith = (params: any) => {
```
TarasHoliuk, line 48: you could use URLSearchParams (same as in `getSearchWith`). It's bad practice to write `any` without necessary

TarasHoliuk (changes requested):
Now loader and placeholder are displayed during loader:

image

You shouldn't display the placeholder in this case

etojeDenys (approved):
good job

PR #867, opened 2024-05-02
Implement filtering & sorting for people table

olafchuszno:
[DEMO LINK] https://olafchuszno.github.io/react_people-table-advanced/


src/components/PeopleFilters.tsx
```diff
@@ -1,18 +1,59 @@
...
+    if (!nameFilter.length) {
+      searchParams.delete('query');
+      setSearchParams(searchParams);
+    } else {
+      searchParams.set('query', nameFilter);
+      setSearchParams(searchParams);
+    }
+  }, [nameFilter, searchParams, setSearchParams]);
+
+  const getCenturies = useCallback(
```
yevhenii-pyl, line 24: No need to memoize it, there is no need to think about referential equality or any complex computations. 
As the discussion point: you are passing an array as the dependency. Will it always work as expected? :)

src/components/PeopleFilters.tsx
```diff
@@ -1,18 +1,59 @@
...
+          className={classNames({ 'is-active': sex === 'm' })}
+          params={{ sex: 'm' }}
+        >
           Male
-        </a>
-        <a className="" href="#/people?sex=f">
+        </SearchLink>
+        <SearchLink
+          className={classNames({ 'is-active': sex === 'f' })}
+          params={{ sex: 'f' }}
```
yevhenii-pyl, line 53: Since you are working with TS, will be better to create a simple enum for genders to avoid somehow 'magical' null | 'f' | 'm' values.

src/components/PeopleFilters.tsx
```diff
@@ -33,63 +78,42 @@ export const PeopleFilters = () => {
...
-            </a>
-
-            <a
-              data-cy="century"
-              className="button mr-1"
-              href="#/people?centuries=20"
-            >
-              20
-            </a>
+            {['16', '17', '18', '19', '20'].map((century: string) => (
```
yevhenii-pyl, line 81: Definitely not a problem, but would be better to create a constant for it. Who knows, maybe, this list of centuries will require adjustments or will be used somewhere else.

src/components/PeopleFilters.tsx
```diff
@@ -33,63 +78,42 @@ export const PeopleFilters = () => {
...
+              <SearchLink
+                params={{ centuries: getCenturies(century) }}
+                className={classNames('button', 'mr-1', {
+                  'is-info': !getCenturies(century).includes(century),
```
yevhenii-pyl, line 85: You are using this function twice in the JSX. Why don't you create a const for it? 
Like `const selectedCenturies = getCenturies(century)` or something.

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,102 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [hasLoadingError, setHasLoadingError] = useState<boolean>(false);
+  const [searchParams] = useSearchParams();
+
+  const query = searchParams.get('query');
+  const centuries = searchParams.getAll('centuries');
+  const sex = searchParams.get('sex');
+  const sortBy = searchParams.get('sort');
+  const order = searchParams.get('order');
+  //   // eslint-disable-next-line no-console
```
yevhenii-pyl, line 23: If there is no need in that code - better to clean it up. If you plan on using this in future, please, drop a comment at the code with a short explanation why is it here.

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,102 @@
 import { PeopleFilters } from './PeopleFilters';
```
yevhenii-pyl, line 1: Please, don't forget about imports order. May seem like a tiny thing, but it all sums up at the end :) 
It usually starts with third-party libs, then - more-less depends on a project, but still, there should be a consistent structure, like: 

```
// third-party, starting from react 
import React from 'react'
...
//API
...
//types
...
// components 
```

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,19 @@
+import classNames from 'classnames';
+import { Person } from '../types';
+
+type Props = {
+  person: Person;
+};
+
+export const PersonLink: React.FC<Props> = ({ person }) => {
+  const { name, slug, sex } = person;
```
yevhenii-pyl, line 9: 👍

src/components/PersonRow.tsx
```diff
@@ -0,0 +1,45 @@
...
+type Props = {
+  person: Person;
+};
+
+export const PersonRow: React.FC<Props> = ({ person }) => {
+  const { sex, born, died, fatherName, motherName, mother, father } = person;
+
+  const { personSlug } = useParams();
+
+  const isChosen = person.slug === personSlug;
```
yevhenii-pyl, line 15: 👍

src/utils/getSortedPeople.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const getSortedPeople = (
+  people: Person[],
+  sortBy: string | null,
+  order: string | null,
+) => {
+  const sortedPeople = people;
+
+  if (sortBy) {
+    if (sortBy) {
```
yevhenii-pyl, line 11: Twice the same condition? Looks like a typo :)

You can avoid any if statement here by just providing a `default` case for switch statement (btw, you should add a default case anyways).

src/utils/getSortedPeople.ts
```diff
@@ -0,0 +1,42 @@
...
+      switch (sortBy) {
+        case 'name':
```
yevhenii-pyl, line 13: And let's create enums for sorting cases, just... in case :)

yevhenii-pyl (changes requested):
Solid work! 🥇 
Let's provide a few small adjustments and we are good to go!

yevhenii-pyl (approved):
You got it 👍

PR #874, opened 2024-05-08
Add task solution

TetianaBukhenko:
[DEMO LINK](https://TetianaBukhenko.github.io/react_people-table-advan
image
ced/)


anastasiiavorobiova (changes requested):
Great work! Consider keeping the sort and filters when the person is changed

VitaliyBondarenko1982 (approved):
Well done

PR #880, opened 2024-05-14
Solution

Davwidos:
[DEMO LINK](https://davwidos.github.io/react_people-table-advanced/)

### DIFF
--- a/src/providers/PeopleProvider.tsx
+++ b/src/providers/PeopleProvider.tsx
@@ -1,0 +1,122 @@
+import {
+  FC,
+  PropsWithChildren,
+  createContext,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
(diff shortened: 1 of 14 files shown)


src/providers/PeopleProvider.tsx
```diff
@@ -0,0 +1,122 @@
...
+  useMemo,
+  useState,
+} from 'react';
+import { Person } from '../types';
+import { getPeople } from '../api';
+import { useParams } from 'react-router-dom';
+import { usePeopleParams } from '../hooks/usePeopleParams';
+import { getCentaury } from '../utils/peopleUtils';
+
+interface PeopleContextI {
```
choeqq, line 16: if you want to use `I` inside your interface name we typically use it as a prefix 😉 
```suggestion
interface IPeopleContext {
```

choeqq (approved):

PR #895, opened 2024-05-24
Solution

Allanpt:
[DEMO LINK](https://Allanpt.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,18 +1,77 @@
-export const PeopleFilters = () => {
+import React from 'react';
+import { Link, SetURLSearchParams } from 'react-router-dom';
+import { SearchParams, getSearchWith } from '../utils/searchHelper';
+import { SearchLink } from './SearchLink';
+
+type Props = {
+  searchParams: URLSearchParams;
+  setSearchParams: SetURLSearchParams;
+  sex: string;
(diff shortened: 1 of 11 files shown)


src/components/PeopleFilters.tsx
```diff
@@ -33,63 +94,68 @@ export const PeopleFilters = () => {
...
+            </Link>
 
-            <a
+            <Link
               data-cy="century"
-              className="button mr-1 is-info"
-              href="#/people?centuries=19"
+              className={`button mr-1 ${centuries.includes('19') ? 'is-info' : ''}`}
+              to="#"
+              onClick={e => handleCenturyClick(e, '19')}
```
jv-aquino, line 128: can you think of a way to simplify this handle here to only receive the event? (hint: ids)

jv-aquino (changes requested):
Hey Allan, good job, for real! Your project is awesome, this will be a crucial part of portfolio, for sure! There are only some small changes that I'd wish you consider (not spoiling anything, just maybe helping you visualize improvements):

- remember the uses of useCallback and useMemo
- remember the attributes of the event object
- more specific and **optional** suggestion: when the user is trying to access the mother/father of the row, redirect it and change the permissions if they're not allowed to 
(e.g. the user selected "Male" and tries to access the mother of someone)

PR #898, opened 2024-05-26
Task solution

kinqbert:
[DEMO LINK](https://kinqbert.github.io/react_people-table-advanced/)


metrakonon (approved):
Great work! But here are some major issues with the filters, take a look.
1. The "All" button should have the class "is-outlined" if no century is selected, as in the working version.
2.  If you select any century, then gender, and press "Reset All Filters," the list should reset to its default state instead of becoming empty.
3. Also, you can check that your results don't match with the working version (one example below).
Screenshot from 2024-05-27 16-37-16
Screenshot from 2024-05-27 16-37-03

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,211 @@
...
+      !person.motherName?.includes(query) &&
+      !person.fatherName?.includes(query)
+    ) {
+      return false;
+    }
+
+    // checks if born or death year is within certain century
+    if (
+      centuries &&
+      !centuries.includes((person.born % 100).toString()) &&
```
metrakonon, line 62: I believe there should be something like this: `centuries.includes(${Math.ceil(person.born` / `100)})`

kinqbert:
@metrakonon Huge thanks! Didn't catch those bugs while was doing the task. Implemented fixes into my code. Appreciate it once again! 🤝

src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <nav
+      data-cy="nav"
+      className="navbar is-fixed-top has-shadow"
+      role="navigation"
+      aria-label="main navigation"
+    >
+      <div className="container">
+        <div className="navbar-brand">
+          <NavLink className={getElementClass} to="./" end>
```
DanilWeda, line 17: you can use an enum for all paths and use them in all places in your application

DanilWeda (approved):
Good job! 
Approved with a small comment.
Thanks)

PR #904, opened 2024-05-29
task solution

MatheusStefan:
[DEMO LINK](https://MatheusStefan.github.io/react_people-table-advanced/)

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -1,0 +1,20 @@
+import { HashRouter, Navigate, Route, Routes } from 'react-router-dom';
+import { App } from './App';
+import { PeoplePage } from './components/PeoplePage';
+
+export const Root = () => (
+  <HashRouter>
+    <Routes>
+      <Route path="/" element={<App />}>
+        <Route index element={<h1 className="title">Home Page</h1>} />
+        <Route path="home" element={<Navigate to="/" />} />
+
+        <Route path="/people">
+          <Route index element={<PeoplePage />} />
+          <Route path=":slug" element={<PeoplePage />} />
+        </Route>
+        <Route path="*" element={<h1 className="title">Page not found</h1>} />
+      </Route>
+    </Routes>
+  </HashRouter>
+);
--- a/src/components/PeoplePage.tsx
+++ b/src/components/PeoplePage.tsx
@@ -1,8 +1,125 @@
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { useEffect, useMemo, useState } from 'react';
+import { Person } from '../types';
+import { getPeople } from '../api';
+import { useParams, useSearchParams } from 'react-router-dom';
+import { PeopleMap } from '../types/PeopleMap';
+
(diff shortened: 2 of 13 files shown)


src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,163 @@
...
+        return sortDirection ? 1 : -1;
+      }
+
+      if (valueA > valueB) {
+        return sortDirection ? -1 : 1;
+      }
+
+      return 0;
+    });
+  }, [filteredPeople, sortColumn, sortDirection]);
```
pedro-ruas, line 95: Add a new line here, would ya?

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,163 @@
...
+  const handleSearchChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const newParams = new URLSearchParams(searchParams);
+
+    newParams.set('query', event.target.value);
+    setSearchParams(newParams);
```
pedro-ruas, line 100: Not necessary, but maybe you could use something similar to the approach in setState:
```
setSearchParams(params => {
    params.set('query', event.target.value);
    return params;
});

File: src/components/PeoplePage.tsx [hunk 3]
```diff
@@ -1,29 +1,163 @@
...
     <>
       <h1 className="title">People Page</h1>
 
       <div className="block">
         <div className="columns is-desktop is-flex-direction-row-reverse">
           <div className="column is-7-tablet is-narrow-desktop">
-            <PeopleFilters />
+            <PeopleFilters
+              onSearchChange={handleSearchChange}
+              sex={sex ?? undefined}
```
pedro-ruas, line 135: What value would _sex_, _centuries_ and _query_ have if when trying to declare them, _searchParams_ didn't have the key you're loking for?
E.g:
```
const sex = searchParam.get('sex');
    // if searchParams = { query: 'batata', centuries: ['15'] }; What would be the value of _sex_ ? 
```

pedro-ruas (changes requested):
Well done! Some improvements only

src/Root.tsx
```diff
@@ -0,0 +1,20 @@
...
+  <HashRouter>
+    <Routes>
+      <Route path="/" element={<App />}>
+        <Route index element={<h1 className="title">Home Page</h1>} />
+        <Route path="home" element={<Navigate to="/" />} />
+
+        <Route path="/people">
+          <Route index element={<PeoplePage />} />
+          <Route path=":slug" element={<PeoplePage />} />
+        </Route>
```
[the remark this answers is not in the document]
MatheusStefan: Unfortunately I can't modify that way, cause it needs to have a param "people" and then another param ":slug" as shown here

If i keep my code code, the people page renders normally:
imagem_2024-05-29_210642788

If I use in the same route, the people page doesn't render:
imagem_2024-05-29_210919437

So, the slug needs to be a subroute of people.

EdPirro (approved):
Great job, @MatheusStefan! Keep it up! 🚀

Please review the last comment @pedro-ruas left (I mentioned you on it). But, other than that detail, your code looks great!

PR #925, opened 2024-06-03
Solution

dyakym:
[DEMO LINK](https://dyakym.github.io/react_people-table-advanced/)


src/components/PeopleLink/PeopleLink.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+type PeopleLinkProps = {
+  person: Person;
+};
+
+export const PeopleLink: React.FC<PeopleLinkProps> = ({ person }) => {
+  const { name, sex, born, died, motherName, fatherName, mother, father } =
+    person;
+  const { slug } = useParams();
+  const female = sex === 'f';
```
undeletable, line 13: Better to use `isFemale` name - to indicate that it's boolean. Usage of `female` name feels like it might be some string/object/etc.

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,62 @@
...
+    setIsLoading(true);
+    getPeople()
+      .then(setPeople)
+      .catch(() => setError('Something went wrong'))
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const [searchParams] = useSearchParams();
+  const newPeople = getFilteredPeople(people, searchParams);
+  const wrongData = ! people.length && !isLoading && !error;
```
undeletable, line 25: `isWrongData` is more appropriate name, which implies that it's boolean. Usage of `wrongData` name implies that it's some kind of data.

undeletable (changes requested):
The solution looks fine. Just please fix eslint warnings - there are a lot of them, and also take a look at some namings. Thanks.

PR #943, opened 2024-06-21
Solution

sevenlays:
[DEMO LINK](https://sevenlays.github.io/react_people-table-advanced/)


src/components/PersonList/PersonList.tsx
```diff
@@ -0,0 +1,128 @@
...
+          <i className="fas fa-sort-down" />
+        </span>
+      );
+    }
+
+    return (
+      <span className="icon">
+        <i className="fas fa-sort" />
+      </span>
+    );
```
DanilWeda, line 48: You can optimize this element duplication. 
Try to write func which returns your ready class ( vector). 
Then return the element with this class. 

```suggestion
    return (
      <span className="icon">
        <i className=`fas ${vector}` />
      </span>
    );
```

src/helpers/useChangeSearchParams.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  const setSearhParams = (prop: TableTitles) => {
+    const params = new URLSearchParams(searchParams);
+
+    if (!column) {
+      params.set('sort', prop);
+    }
+
+    if (column === prop) {
+      params.set('order', 'desc');
```
DanilWeda, line 18: In this case also use enum

DanilWeda (changes requested):
Good job. 
Please check some comments which I left to improve your solution. 
Many thanks)

mrkovaliv (approved):
👍 👍 👍

PR #963, opened 2024-07-11
Solution

ivankovbohdan:
[DEMO LINK](https://ivankovbohdan.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeoplePage.tsx
+++ b/src/components/PeoplePage.tsx
@@ -1,8 +1,130 @@
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { getPeople } from '../api';
+import { useEffect, useState } from 'react';
+import { Person } from '../types';
+import { useParams, useSearchParams } from 'react-router-dom';
+
+function preperedPeople(
+  query: string,
+  filterField: string,
+  centuries: string[],
+  sortField: string,
+  sortOrder: string,
+  readyPeople: Person[],
+) {
+  let visiblePeople = [...readyPeople];
+  const preperedQuery = query.trim().toLowerCase();
+
+  if (sortField) {
+    visiblePeople.sort((a: Person, b: Person) => {
+      switch (sortField) {
+        case 'name':
+          if (sortOrder === 'asc') {
+            return a.name.localeCompare(b.name);
+          } else if (sortOrder === 'desc') {
+            return b.name.localeCompare(a.name);
+          } else {
+            return 0;
+          }
+
+        case 'sex':
+          if (sortOrder === 'asc') {
+            return b.sex.localeCompare(a.sex);
+          } else if (sortOrder === 'desc') {
+            return a.sex.localeCompare(b.sex);
+          } else {
+            return 0;
+          }
+
+        case 'born':
+          if (sortOrder === 'asc') {
+            return a.born - b.born;
+          } else if (sortOrder === 'desc') {
+            return b.born - a.born;
+          } else {
+            return 0;
+          }
+
+        case 'died':
+          if (sortOrder === 'asc') {
+            return a.died - b.died;
+          } else if (sortOrder === 'desc') {
+            return b.died - a.died;
+          } else {
+            return 0;
+          }
+
+        default:
+          return 0;
+      }
+    });
+  }
+
+  visiblePeople = visiblePeople.filter(
+    person =>
+      person.name.toLowerCase().includes(preperedQuery) ||
+      person.motherName?.toLowerCase().includes(preperedQuery) ||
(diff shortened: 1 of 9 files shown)


src/components/PeopleTable.tsx
```diff
@@ -21,7 +75,14 @@ export const PeopleTable = () => {
           <th>
             <span className="is-flex is-flex-wrap-nowrap">
               Sex
-              <a href="#/people?sort=sex">
+              <a
```
TarasHoliuk, line 78: You should use the Link component from react-router here as well
Moreover, in this task, you have already created SearchLink that will be useful here

src/components/PeopleTable.tsx
```diff
@@ -57,588 +132,44 @@ export const PeopleTable = () => {
...
+                  className="has-text-danger"
+                >
+                  {person.motherName}
+                </Link>
+              ) : (
+                person.motherName || '-'
+              )}
+            </td>
+            <td>
+              {person.father ? (
```
TarasHoliuk, line 163: Why not PersonLink here? Same for mother

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,172 @@
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { getPeople } from '../api';
+import { useEffect, useState } from 'react';
+import { Person } from '../types';
+import { useParams, useSearchParams } from 'react-router-dom';
+
+function preperedPeople(
```
TarasHoliuk, line 9:
```suggestion
function getPreperedPeople(
```
it's a function - preparedPeople sounds like it's already an array of prepared people

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,172 @@
...
+          }
+
+        case 'sex':
+          if (sortOrder === 'asc') {
+            return b.sex.localeCompare(a.sex);
+          } else if (sortOrder === 'desc') {
+            return a.sex.localeCompare(b.sex);
+          } else {
+            return 0;
+          }
```
TarasHoliuk, line 39: logic for string values are the same in different cases
You can join cases:

```
case 'name':
case 'sex':
  // logic here b[sortField].localCompare
```
Same for numeric fields

TarasHoliuk (changes requested):
GJ 👍 
Let's add a few fixes

src/components/PeopleTable.tsx
```diff
@@ -10,44 +57,72 @@ export const PeopleTable = () => {
           <th>
             <span className="is-flex is-flex-wrap-nowrap">
               Name
-              <a href="#/people?sort=name">
+              <SearchLink
+                params={{ sort: 'name' }}
+                onClick={e => {
```
TarasHoliuk, line 62: instead of changing params manually on click you should just set the correct params for link
You can use function to get correct params (including order) based on current searchParams values

That's the way we want to work with Router links 🙂 So after that remove onClick handler at all

Also, use Search link in PeopleFilters

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Person } from '../types';
+
+type Props = {
+  person: Person | undefined;
+};
+
+export const PersonLink: React.FC<Props> = ({ person }) => (
+  <Link
+    to={`/people/${person?.slug}`}
+    className={classNames('', {
```
TarasHoliuk, line 12:
```suggestion
    className={classNames({
```

TarasHoliuk (changes requested):
Good progress! Let's improve Router link usage 🙂

etojeDenys (changes requested):
1. do not show this message on loading
image

2. you need to change the icon on click
image

3. looks different then on working example
image

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,156 @@
...
+  switch (filterField) {
+    case 'all':
+      return visiblePeople;
+    case 'm':
+      return visiblePeople.filter(person => person.sex === 'm');
+    case 'f':
+      return visiblePeople.filter(person => person.sex === 'f');
+    default:
+      return visiblePeople;
+  }
```
etojeDenys, line 72:
```suggestion

  if (filterField) {
     visiblePeople = visiblePeople.filter(person => person.sex === filterField);
  }
  
  return visiblePeople;
```
we can simplify that

etojeDenys (changes requested):
https://github.com/user-attachments/assets/79e86a1a-97dc-4ceb-afca-83356d5cfc01

do not reset the query params on click

VitaliyBondarenko1982 (approved):
Looks good

PR #966, opened 2024-07-15
add solution

Nazarin565:
[DEMO LINK](https://Nazarin565.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeoplePage.tsx
+++ b/src/components/PeoplePage.tsx
@@ -1,7 +1,7 @@
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { getPeople } from '../api';
 import { Person } from '../types';
 import { FilterBySex } from '../types/FilterBySex';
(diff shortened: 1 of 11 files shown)


src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,112 @@
...
+  const [searchParams] = useSearchParams();
+  const sex = searchParams.get('sex') || '';
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+
+  const [activeSexFilter, setActiveSexFilter] = useState<FilterBySex>(
+    FilterBySex.ALL,
+  );
+
+  useEffect(() => {
```
TarasHoliuk, line 24: in such cases you don't need a separate state for `activeSexFilter`
It can be replaced with common variable (wrapped in useMemo):
```
const activeSexFilter = useMemo(() => {
  if (sex === 'm') {
    return FilterBySex.MALE;
  } 
  
  if (sex === 'f') {
    return FilterBySex.FEMALE;
  } 

   return FilterBySex.ALL;
}, [sex]);
```

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,112 @@
...
+  }, [sex]);
+
+  useEffect(() => {
+    getPeople()
+      .then(setPeople)
+      .catch(() => setIsError(true))
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const visibleList = () => {
```
TarasHoliuk, line 41: if it's a function it should be named accordingly: `getVisibleList`. Also, in such case, it's better to move it outside of the component and accept all required  data as variables 

Otherwise (better in this case),  you can just use useMemo hook to do the same thing and memorize value

Probably, you are not familiar with useMemo yet. In such case you can apply changes without using this hook (or just read about it if you want 🙂 )

src/components/PeoplePage.tsx
TarasHoliuk: General rule - do not create separate state if you can calculate it using other state/props

TarasHoliuk (changes requested):
Nicely done
Added a few small suggestions

TarasHoliuk (approved):
✔️

PR #984, opened 2024-07-22
Solution

Kasprukov:


src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,109 @@
...
+  }, []);
+
+  const visiblePeople = useMemo(() => {
+    let visible = [...people];
+
+    if (sex) {
+      switch (actualSexFilter) {
+        case FilterBySex.MALE:
+          visible = visible.filter(person => person.sex === 'm');
+          break;
```
OSpyrydonov, line 45: 'm' is magic value, create const for this

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,109 @@
...
+        case FilterBySex.FEMALE:
+          visible = visible.filter(person => person.sex === 'f');
+          break;
```
OSpyrydonov, line 48: 'f' also magic value

OSpyrydonov (approved):
Great Work !

PR #996, opened 2024-08-07
add task solution

V-Shut:


priemssonia (approved):
Well done!! 😍

artemforsoff (changes requested):
_Implement sorting by name, sex, born and died by clicking on arrows in a th_

sorting parameters in URL are saved but they do not affect and do not work

Moroz-Dmytro (changes requested):
When I switch from `Male` to `Female` all other filters should not be reseted
When I click on link in table, filters should not be reseted

PR #1004, opened 2024-08-08
add initial solution

lesdyman:
Hi! Here is my initial solution for the react_people-table-advanced task. It could be a little bit chaotic as I started solving it before group project and finished just now. And also I had some annoying enviromment problems.  Anyway you shall find demo link below as always. Looking forward for advices how to make this dr. Frankenstein's monster more alive! Cheers!

[DEMO] (https://lesdyman.github.io/react_people-table-advanced/)


HumenVAsya (approved):
everything works, that's the main thing)

PR #1007, opened 2024-08-09
my solution

kostiukmkalne:
https://kostiukmkalne.github.io/react_people-table-advanced/


etojeDenys (changes requested):
your demo link doesn't work, try to fix that

PR #1018, opened 2024-08-19
add task solution

VladKugot:
[DEMO LINK](https://VladKugot.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeopleTable.tsx
+++ b/src/components/PeopleTable.tsx
@@ -1,5 +1,28 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
-export const PeopleTable = () => {
+import classNames from 'classnames';
+import { Link, useParams } from 'react-router-dom';
+import { Person } from '../types';
+import { FC } from 'react';
+import { SortLink } from './SortLink';
+
+enum Sex {
+  Female = 'f',
+  Male = 'm',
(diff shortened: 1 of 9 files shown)


src/components/PeopleTable.tsx
```diff
@@ -10,635 +33,90 @@ export const PeopleTable = () => {
...
+                  >
+                    {motherName}
+                  </Link>
+                ) : (
+                  <span>{motherName || '-'}</span>
+                )}
+              </td>
+              <td>
+                {fatherName && findPersonByName(fatherName)?.slug ? (
+                  <Link to={`/people/${findPersonByName(fatherName)?.slug}`}>
```
etojeDenys, line 110: you can save the result of the `findPersonByName(fatherName)` function to some var and reuse it everywhere instead of calling it twice

etojeDenys (changes requested):
https://github.com/user-attachments/assets/619359a7-f0ad-437c-9730-90bd8016498d

do not reset filters on the person click

etojeDenys (changes requested):
still not fixed, don't hesitate to ask for some help in the chat

https://github.com/user-attachments/assets/a7cdd29e-852a-4710-b183-e0835700faa9

PR #1020, opened 2024-08-19
task solution

BoGoBooM:
[DEMO LINK](https://bogoboom.github.io/react_people-table-advanced/)


src/App.tsx
```diff
@@ -10,9 +10,7 @@ export const App = () => {
 
       <div className="section">
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Outlet />
```
TarasHoliuk, line 13: You should properly style NavLinks in the Navbar:
image

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,116 @@
...
+  filterPersonByQuery,
+  getPeopleWithParents,
+  getSortPeople,
+} from '../utils/functions';
+import { SortBy } from '../types/SortBy';
+import { PeopleFilters } from '../components/PeopleFilters';
+
+export const PeoplePage: React.FC = () => {
+  const [peopleFromServer, setPeopleFromServer] = useState<Person[]>([]);
+  const [isError, setIsError] = useState('');
```
TarasHoliuk, line 18: if you store error here do not use `is` in the start. Use `is`, `has` for boolean variables
 ```suggestion
  const [error, setError] = useState('');
```

File: src/types/FilterByCentury.ts [hunk 3]
```diff
@@ -0,0 +1,7 @@
+export enum FilterByCentury {
```
TarasHoliuk, line 1: In this case union type would fit better, at least you can try to use it)

TarasHoliuk (changes requested):
GJ 👍 
Let's fix a few small things

Also, fix all tests

src/components/PersonInfo.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Person } from '../types';
+
+type Props = {
+  person: Person;
+  search: URLSearchParams;
+};
+
+export const PersonInfo: React.FC<Props> = ({ person, search }) => {
+  const { name, slug, sex } = person;
+  const SEX_FEMALE = 'f';
```
volodymyr-soltys97, line 12: Don't need to create a variable for 'f', already you have a enum 
image

volodymyr-soltys97 (changes requested):
Hi, you need to fix this failed test, if you need help feel free in the fe_chat 
image
image

Add styles here for active filter 
image

src/utils/functions.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export const getSortPeople = (people: Person[], sortBy: SortBy): Person[] => {
+  return [...people].sort((per1, per2) => {
+    if (sortBy === SortBy.Name || sortBy === SortBy.Sex) {
+      return per1[sortBy].localeCompare(per2[sortBy]);
+    }
+
+    if (sortBy === SortBy.Born || sortBy === SortBy.Died) {
+      return per1[sortBy] - per2[sortBy];
+    }
```
etojeDenys, line 68: better to use switch case here instead of several if/else

PR #1022, opened 2024-08-20
add solution

marichkamt:
[DEMO LINK](https://marichkamt.github.io/react_people-table-advanced/)

### DIFF
--- a/src/utils/filterPeople.ts
+++ b/src/utils/filterPeople.ts
@@ -1,0 +1,70 @@
+import { Person } from '../types/Person';
+
+export function filterPeopleBySearchParams(
+  peopleToFilter: Person[],
+  params: URLSearchParams,
+): Person[] {
+  const sex = params.get('sex');
+  const query = params.get('query');
+  const centuries = params.getAll('centuries');
+  const sort = params.get('sort');
+  const order = params.get('order');
+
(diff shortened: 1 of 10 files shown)


src/utils/filterPeople.ts
```diff
@@ -0,0 +1,70 @@
...
+      case 'name':
+        people.sort((person1, person2) =>
+          person1.name.localeCompare(person2.name),
+        );
+        break;
+      case 'sex':
+        people.sort((person1, person2) =>
+          person1.sex.localeCompare(person2.sex),
+        );
+        break;
```
etojeDenys, line 54:
```suggestion
      case 'name':
      case 'sex':
        people.sort((person1, person2) =>
          person1[sort].localeCompare(person2[sort]),
        );
        break;
```
we can group some cases

src/utils/filterPeople.ts
```diff
@@ -0,0 +1,70 @@
...
+      case 'born':
+        people.sort((person1, person2) => person1.born - person2.born);
+        break;
+      case 'died':
+        people.sort((person1, person2) => person1.died - person2.died);
+        break;
```
etojeDenys, line 60: same here

etojeDenys (changes requested):
looks different in the working example
image

PR #1026, opened 2024-08-25
Sorted

LabPetro:
[DEMO LINK](https://LabPetro.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeoplePage.tsx
+++ b/src/components/PeoplePage.tsx
@@ -1,8 +1,24 @@
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { useEffect, useState } from 'react';
+import { Person } from '../types';
+import { getPeople } from '../api';
+import { Message } from '../enums/Message';
 
 export const PeoplePage = () => {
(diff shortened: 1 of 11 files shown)


src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,49 @@
...
-              <p data-cy="peopleLoadingError">Something went wrong</p>
+              {error && <p data-cy="peopleLoadingError">{Message.Error}</p>}
 
-              <p data-cy="noPeopleMessage">There are no people on the server</p>
+              {!isLoading && !error && !people.length && (
+                <p data-cy="noPeopleMessage">{Message.NoPeople}</p>
+              )}
 
-              <p>There are no people matching the current search criteria</p>
+              {false && <p>{Message.Missmatch}</p>}
```
natalia-klonowska, line 42: you should use a condition instead of false because right now this message will never appear

natalia-klonowska (changes requested):
century filter is not working:
image
query should not only look for match in name but also in motherName and fatherName:
image
[working example:](https://mate-academy.github.io/react_people-table-advanced/)
image
yours: 
image

natalia-klonowska (approved):
Great work 🚀

PR #1042, opened 2024-09-05
solution

flocky674:
[DEMO LINK] - https://flocky674.github.io/react_people-table-advanced/


etojeDenys (changes requested):
you should change those arrows on click
image
let's fix that firstly 

and why do you need this section here

PR #1050, opened 2024-09-08
add task solution

Follder:


Follder:
[DEMO LINK](https://follder.github.io/react_people-table-advanced/)

PR #1053, opened 2024-09-09
Solution

opalahecha:
[DEMO LINK](https://opalahecha.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/PeoplePage.tsx
+++ b/src/components/PeoplePage.tsx
@@ -1,8 +1,141 @@
+import { useSearchParams } from 'react-router-dom';
+import React, { useEffect, useState } from 'react';
 import { PeopleFilters } from './PeopleFilters';
 import { Loader } from './Loader';
 import { PeopleTable } from './PeopleTable';
+import { getPeople } from '../api';
+import { getSearchWith } from '../utils/searchHelper';
+import { Person } from '../types';
 
 export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [hasError, setHasError] = useState(false);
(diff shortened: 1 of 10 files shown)


src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,172 @@
...
+    const matchesQuery = person.name.toLowerCase().includes(query);
+    const matchesCentury =
+      centuries.length === 0 ||
+      centuries.includes(Math.ceil(person.born / 100));
+    const matchesSex = !sexFilter || person.sex === sexFilter;
+
+    return matchesQuery && matchesCentury && matchesSex;
+  });
+
+  const sortedPeople = [...filteredPeople].sort((a, b) => {
```
TarasHoliuk, line 89: You can wrap it into useMemo to avoid recalculating it every time an unrelated state or search parameter changes. Same for filteredPeople

src/components/personLink.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface PersonLinkProps {
+  person: Person;
+}
+
+export const PersonLink: React.FC<PersonLinkProps> = ({ person }) => {
+  return (
+    <td>
+      <Link
+        to={`./${person.slug}`}
```
TarasHoliuk, line 14: It resets all filters. That unexpected behavior - should be fixed

TarasHoliuk (changes requested):
Great job! A few things should be improved

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-09-10 at 15 40 47
Consider setting the same parameter just once

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1064, opened 2024-09-20
my solution

viktalfrontend:


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. This block should be shown only after people have downloaded it
image

2. Don't need reset filters after clicking on the person 
image
image
image

3. Fix all linter errors 
image

src/components/PeopleFilters.tsx
```diff
@@ -1,16 +1,103 @@
...
+    if (sex) {
+      params.set('sex', sex);
+    } else {
+      params.delete('sex');
+    }
+
+    setSearchParams(params);
+  }
+
+  function handleQueryChange(event: React.ChangeEvent<HTMLInputElement>) {
```
volodymyr-soltys97, line 28: It is important to follow the same way of creating functions, fix it everywhere
```suggestion
  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,122 @@
...
+                </p>
+              )}
 
-              <p>There are no people matching the current search criteria</p>
+              {!loading && !error && !!filteredPeople.length && (
+                <PeopleTable people={filteredPeople} />
+              )}
 
-              <PeopleTable />
+              {!loading && !error && filteredPeople.length === 0 && (
```
volodymyr-soltys97, line 115:
```suggestion
              {!loading && !error && !filteredPeople.length && (
```

volodymyr-soltys97 (changes requested):
Almost done!

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1075, opened 2024-10-05
add task solution

julius0724:
[DEMO LINK](https://julius0724.github.io/react_people-table-advanced/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Fix the filter block position, check it on the working example 
image

Fix the search UI, check it 
image

Don't need to reset filters after clicking on the person

vadiimvooo (approved):
Well done

PR #1084, opened 2024-10-14
Develop

RostyslavSharuiev:
- [DEMO LINK](https://RostyslavSharuiev.github.io/react_people-table-advanced/)


src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,32 @@
...
+      role="navigation"
+      aria-label="main navigation"
+    >
+      <div className="container">
+        <div className="navbar-brand">
+          <NavLink to="/" className={getLinkClass}>
+            Home
+          </NavLink>
+
+          <NavLink to="/people" aria-current="page" className={getLinkClass}>
```
hma-3, line 25: [TASK] Keep search params when navigating within the People page (when selecting a person or **clicking the People link**).

src/components/PeopleFilter/PeopleFilters.tsx
```diff
@@ -0,0 +1,100 @@
...
+      </p>
+
+      <div className="panel-block">
+        <p className="control has-icons-left">
+          <input
+            data-cy="NameFilter"
+            type="search"
+            className="input"
+            placeholder="Search"
+            value={query || ''}
```
hma-3, line 48: why do you need `|| ''` here? you already did this

> const query = searchParams.get('query') || '';

src/components/PeopleFilter/PeopleFilters.tsx
```diff
@@ -0,0 +1,100 @@
...
+              params={{ century: [] }}
+            >
+              All
+            </SearchLink>
+          </div>
+        </div>
+      </div>
+
+      <div className="panel-block">
+        <Link className="button is-link is-outlined is-fullwidth" to="/people">
```
hma-3, line 94: managing the reset in this way is not entirely correct, what if a person was selected?

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { Person } from '../../types';
+
+import { sortTypes } from './../../constants/constants';
+
+import { PersonLink, SearchLink } from '../';
+
+interface Props {
+  people: Person[];
+  sort: string | null;
+  order: string | null;
```
hma-3, line 15: do you really need to get sort and order as props?

src/utils/getVisibleFilter.ts
```diff
@@ -0,0 +1,51 @@
+import { Person } from '../types';
+import { SortTypes } from '../types/SortTypes';
+
+export const getVisiblePeople = (
+  people: Person[],
+  query: string | null,
+  sex: string | null,
+  century: string[],
+  sort: string | null,
+  order: string | null,
```
hma-3, line 10: > do you really need to get sort and order as props?

same

hma-3 (changes requested):
Good job!
But one task wasn't fully realized and also have a few suggestions

PR #1085, opened 2024-10-16
solution

pasha1932:
[DEMO LINK](https://pasha1932.github.io/react_people-table-advanced/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
       },
       "devDependencies": {
         "@cypress/react18": "^2.0.1",
-        "@mate-academy/scripts": "^1.8.5",
+        "@mate-academy/scripts": "^1.9.12",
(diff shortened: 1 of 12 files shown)


etojeDenys (changes requested):
image
you need show the message instead of the table in this case(like in the working example)

2. do not reset filters on click

https://github.com/user-attachments/assets/3a6417c7-2f12-4c3a-b4c7-81d697f5eda9

pasha1932:
the ai comment my solution like this ''The toggleCenturies function is being called immediately here, which means it will execute during the render phase. This is likely not the intended behavior. Consider wrapping the call in a function to ensure it is called only when the SearchLink is interacted with'', i need to change my code or the ai make a mistake?

etojeDenys (approved):

PR #1094, opened 2024-10-28
solution

YuliiaShatkovska:
[DEMO LINK](https://YuliiaShatkovska.github.io/react_people-table-advanced/)


EustasTheMonk (approved):
Overall looks perfect but the only thing i'd suggest is to use `!array.length` instead of `array.length === 0` and  `!!array.length` instead of `array.length > 0`

PR #1095, opened 2024-10-28
temporary solution

Svidruk:
Not ready yet


src/Root.tsx
```diff
@@ -0,0 +1,27 @@
...
+export const Root = () => (
+  <Router>
+    <Routes>
+      <Route path="/" element={<App />}>
+        <Route index element={<HomePage />} />
+        <Route path="*" element={<NotFoundPage />} />
+        <Route path="/home" element={<Navigate to="/" replace />} />
+        <Route path="people">
+          <Route index element={<PeoplePage />} />
+          <Route path=":slug" element={<PeoplePage />} />
```
YuriiMykhailenko, line 22: It's not necessary to duplicate <PeoplePage> here.

YuriiMykhailenko (changes requested):
Almost there👍

PR #1097, opened 2024-10-28
Develop

Hradivnyk:
[DEMO LINK](https://Sholudyvyy.github.io/react_people-table-advanced/)


src/components/FilterByCentury/CenturyButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+import React from 'react';
+import { useSearchParams } from 'react-router-dom';
+
+type Props = {
+  century: string;
+};
+
+export const CenturyButton: React.FC<Props> = ({ century }) => {
+  const [searchParams] = useSearchParams();
+  let paramsCenturies = [...searchParams.getAll('centuries')];
```
Ann-Maksay, line 12: There are more elegant ways to solve this problem than using let to store an array. For example, using a helper function that will return a new array and write it to the reference parameters. Let for an array is almost always not the best idea.

src/utils/clearPeopleList.ts
```diff
@@ -0,0 +1,25 @@
...
+      (person.name.includes(query || '') ||
+        person.motherName?.includes(query || '') ||
+        person.fatherName?.includes(query || '')) &&
+      (!centuries.length ||
+        centuries.includes(Math.ceil(person.born / 100).toString())),
+  );
+
+  const sortedPeople = sortPeople(sort, order, visiblePeople);
+
+  return sortedPeople;
```
Ann-Maksay, line 24:
```suggestion
  return sort ? sortPeople(sort, order, visiblePeople)  :  visiblePeople;
```
Do we really need to call the sort function even if we don't have any sorting parameters?

Ann-Maksay (approved):
Nice work 💫

PR #1099, opened 2024-10-28
Solution

Sukulynt:
[DEMO LINK](https://Sukulynt.github.io/react_people-table-advanced/)


Sholudyvyy (commented):
In the index.html you need to add class="has-navbar-fixed-top". They Will add space on the top 
image

src/components/PeopleTable.tsx
```diff
@@ -1,644 +1,164 @@
...
+    }
+
+    if (gender) {
+      filtered = filtered.filter(person => person.sex === gender);
+    }
+
+    if (centuries.length > 0) {
+      filtered = filtered.filter(person => {
+        const century = Math.floor(person.born / 100) + 1;
+        return centuries.includes(century.toString());
```
Sholudyvyy, line 42: Expected blank line before this statement.

src/components/PeopleTable.tsx
```diff
@@ -1,644 +1,164 @@
...
+    }
+
+    return sorted;
+  }, [filteredPeople, sort, order]);
+
+  const toggleOrder = (sortName: string) => {
+    if (sort === sortName) {
+      return { sort: sortName, order: order === 'asc' ? 'desc' : 'asc' };
+    }
+    return { sort: sortName, order: 'asc' };
```
Sholudyvyy, line 74: Expected blank line before this statement

src/components/PeopleTable.tsx
```diff
@@ -1,644 +1,164 @@
...
+  };
```
Sholudyvyy, line 75: Maybe it will be better to move these functions to utils

src/components/PeopleTable.tsx
Sholudyvyy, line 74: The example page doesn't have order='asc', because it's default and we don't need it in the link. 

The third click (when people are already sorted in reversed order by this field) disables sorting;

Sholudyvyy (approved):
Nice work 👍

PR #1102, opened 2024-10-29
develop

Ostkreuzzz:
https://Ostkreuzzz.github.io/react_people-table-advanced/


src/components/Person/Person.tsx
```diff
@@ -0,0 +1,55 @@
...
+        </NavLink>
+      </td>
+
+      <td>{person.sex}</td>
+      <td>{person.born}</td>
+      <td>{person.died}</td>
+
+      {person.mother ? (
+        <td>
+          <NavLink
```
vikshiyan, line 33: You can use `<Link>`. The NavLink  is used for creating navigation links that can automatically apply styles or classes to the active link based on the current route.

src/components/PeopleTable.tsx
vikshiyan, line 1: I didn't see this file but you have problem with icon.  On desc, the icon should be down
image

src/index.tsx
vikshiyan, line 1: You have problem with class `class="has-navbar-fixed-top"` add this class to your html 
image
image

PR #1104, opened 2024-10-30
Task solution

WatarJoy:
https://WatarJoy.github.io/react_people-table-advanced/

### DIFF
--- a/src/components/PeopleFilters.tsx
+++ b/src/components/PeopleFilters.tsx
@@ -1,18 +1,51 @@
-export const PeopleFilters = () => {
+import React from 'react';
+import cn from 'classnames';
+
+interface Props {
+  sex: string;
+  query: string;
+  centuries: string[];
+  onReset: (param: string) => void;
+  onSexChange: (sexFilter: string) => void;
+  setCenturies: (century: string) => void;
+  onQueryChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+export const PeopleFilters: React.FC<Props> = ({
+  sex,
+  query,
+  centuries,
+  onSexChange,
+  onReset,
+  onQueryChange,
+  setCenturies,
+}) => {
+  const handleResetAll = () => {
+    onReset('sex');
+    onReset('query');
+    onReset('centuries');
+  };
+
+  const centuriesArray = ['16', '17', '18', '19', '20'];
+
   return (
     <nav className="panel">
       <p className="panel-heading">Filters</p>
 
       <p className="panel-tabs" data-cy="SexFilter">
-        <a className="is-active" href="#/people">
-          All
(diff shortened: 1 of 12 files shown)


src/components/PeopleFilters.tsx
```diff
@@ -1,18 +1,51 @@
...
-          Male
-        </a>
-        <a className="" href="#/people?sex=f">
-          Female
-        </a>
+        {[
+          { param: '', title: 'All' },
+          { param: 'm', title: 'Male' },
+          { param: 'f', title: 'Female' },
+        ].map(filter => (
```
v1RnT, line 40: it would've been better if you put that array into constant instead of what you did

```tsx
const filterOptions = [
          { param: '', title: 'All' },
          { param: 'm', title: 'Male' },
          { param: 'f', title: 'Female' },
        ]

File: src/components/PeopleFilters.tsx [hunk 2]
```diff
@@ -33,52 +68,25 @@ export const PeopleFilters = () => {
...
+                key={filter}
+                data-cy="century"
+                className={cn('button mr-1', {
+                  'is-info': centuries.includes(filter),
+                })}
+                onClick={() => setCenturies(filter)}
+              >
+                {filter}
+              </a>
+            ))}
```
v1RnT, line 82: just a little **note**: you are iterating through array of centuries, but for some reason u named them as `filter` instead of for example `century`

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,130 @@
...
+
+  const { slug } = useParams();
+
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const query = searchParams.get('query') || '';
+
+  const sex = searchParams.get('sex') || '';
+
+  const centuries = searchParams.getAll('centuries') || [];
```
v1RnT, line 24: these are the same type of constants, therefore those spaces don't make much sense

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,26 @@
...
+  person: Person;
+}
+
+export const PersonLink: React.FC<Props> = ({ person }) => {
+  const navigate = useNavigate();
+
+  const handleClick = (event: React.MouseEvent) => {
+    event.preventDefault();
+    navigate(`/people/${person.slug}`, { replace: true });
+  };
```
v1RnT, line 15: this behaviour is already included in `Link` component

v1RnT (changes requested):
There's a difference between [working example](https://mate-academy.github.io/react_people-table-advanced/) and your program. When selecting person name, all of your `filters` and `sort` options are reset, as well as when pressing `Reset all filters` button, all `sort` options are reset, which is not the case in working example

PR #1136, opened 2024-11-22
Develop

italomagno:


raulriato (changes requested):
Hey, @italomagno. It seems you forgot to run the `deploy` for your project. Please do it so I can review it

italomagno:
[This is the deployed](https://italomagno.github.io/react_people-table-advanced)

src/components/ErrorComponent.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { ErrorMessages } from './PeoplePage';
+
+interface ErrorComponentProps {
+  message: ErrorMessages;
+}
+
+export function ErrorComponent({ message }: ErrorComponentProps) {
+  return (
+    <>
+      {message === ErrorMessages.SOMETHING_WENT_WRONG && (
```
julianFBarbosa, line 12: The way you implemented this isn't wrong, but i would recommend to use dynamic property access to avoid repetition (and personally i think that's also improves code readability)!

Imagine if we have 50 types of error messages, what would we do?


an example implementation:

```js
    <>
        {message && (
            <p data-cy={`${message}Message`}>
            {ErrorMessages[message]}
            </p>
        )}
    </>
```

src/components/Navbar.tsx
```diff
@@ -8,13 +13,16 @@ export const Navbar = () => {
     >
       <div className="container">
         <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
+          <a
+            className={`navbar-item ${path.includes('home') || path === '/' ? 'has-background-grey-lighter' : ''}`}
```
julianFBarbosa, line 17:
```suggestion
            className={`navbar-item ${path.includes('home') || path === '/' ? 'has-background-grey-lighter' : ''}`}            
```

We can extract it to a variable (or a function) to avoid complexity in the JSX =)

```tsx
    const getHomeLinkClassNames = () => {
        const isHome = path.includes('home') || path === '/'

        return isHome ? 'has-background-grey-lighter' : ''
    }

    ...

    <a className={getHomeLinkClassNames()}>
```

src/components/PeopleFilters.tsx
```diff
@@ -1,18 +1,72 @@
+import React from 'react';
+import { Link, useLocation, useSearchParams } from 'react-router-dom';
+import { getSearchWith, SearchParams } from '../utils/searchHelper';
+
 export const PeopleFilters = () => {
+  const availableCenturies = [16, 17, 18, 19, 20];
```
julianFBarbosa, line 6:
```suggestion
  const availableCenturies = [16, 17, 18, 19, 20];
```

We can move this variable to outside of the component, since it won't change its values anywhere in the code flow, so we can avoid its re-creation across re-renders =)

julianFBarbosa, line 6:
```suggestion
  const availableCenturies = [16, 17, 18, 19, 20];
```

We can move this variable to outside of the component, since we aren't changing it's values anywhere in the code flow, so we can avoid it's re-creation across re-renders =)

julianFBarbosa (changes requested):
I left some comments and suggestions.

But the main target is well done: the app is working the same way as the example! 

Congrats =)

julianFBarbosa (approved):
nice job =)

PR #1160, opened 2024-12-27
Solution

MakksymSly:
[DEMO LINK](https://makksymSly.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/Navbar.tsx
+++ b/src/components/Navbar.tsx
@@ -1,4 +1,10 @@
+import { NavLink } from 'react-router-dom';
+import cn from 'classnames';
 export const Navbar = () => {
+  const isTabActive = ({ isActive }: { isActive: boolean }) => {
+    return isActive ? 'has-background-grey-lighter' : '';
+  };
+
   return (
     <nav
       data-cy="nav"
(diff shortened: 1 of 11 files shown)


src/components/Navbar.tsx
```diff
@@ -8,17 +14,24 @@ export const Navbar = () => {
...
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
+            className={({ isActive }) =>
+              cn('navbar-item', isTabActive({ isActive }))
+            }
+            to="people"
           >
             People
-          </a>
+          </NavLink>
```
taniabarkovskya, line 34: here you forgot to add this. make sure to do this here and in all People link in table

> "Keep search params when navigating within the People page (when selecting a person or clicking the People link)."

src/components/PeopleFilters.tsx
```diff
@@ -1,18 +1,122 @@
...
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const query = searchParams.get('query') || '';
+
+  const handleQuery = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const params = new URLSearchParams(searchParams);
+
+    params.set('query', event.target.value);
+    setSearchParams(params);
+  };
```
taniabarkovskya, line 15: you need to make sure, that if query is empty, it should be deleted from searchParams

src/components/PeopleFilters.tsx
```diff
@@ -1,18 +1,122 @@
...
+  const handleGenderChange = (
+    gender: Sex,
+    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+  ) => {
+    event.preventDefault();
+    const params = new URLSearchParams(searchParams);
+
+    params.set('sex', gender);
+    setSearchParams(params);
+  };
```
taniabarkovskya, line 26: and for sex as well as for query - empty parameter should not be in search params

taniabarkovskya (changes requested):
you did good job and interesting solution ( I really like it :) )
but please fix some task requirements and make it perfect! 🚀

PR #1164, opened 2024-12-28
solution

minimal005:
[DEMO LINK](https://minimal005.github.io/react_people-table-advanced/)


src/components/PeopleFilters.tsx
```diff
@@ -1,23 +1,58 @@
...
+  };
+
   return (
     <nav className="panel">
       <p className="panel-heading">Filters</p>
 
       <p className="panel-tabs" data-cy="SexFilter">
-        <a className="is-active" href="#/people">
+        <SearchLink
+          className={cn({ 'is-active': sex === 'All' })}
```
StasSokolov1, line 31: Let's move `All` to const or enum and reuse, I see at least few places where we can use it

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export const PersonLink: React.FC<Props> = ({ person }) => {
+  const { slug, sex, name } = person;
+  const { search } = useLocation();
+
+  return (
+    <Link
+      to={{ pathname: `/people/${slug}`, search }}
+      className={cn({ 'has-text-danger': sex === 'f' })}
```
StasSokolov1, line 17: You already have enum with `f` value, please use here also

src/utils/peopleHelper.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (sex) {
+    peopleList = peopleList.filter(person => person.sex === sex);
+  }
+
+  const normalizeQuery = query.toLowerCase().trim();
+
+  if (normalizeQuery) {
+    peopleList = peopleList.filter(person => {
+      return (
+        person.name.toLowerCase().includes(normalizeQuery) ||
```
StasSokolov1, line 36: `.toLowerCase().includes(normalizeQuery)` we can move it to separated function and reuse

StasSokolov1 (changes requested):
Just a few minor comments, let's fix it before approving!

StasSokolov1 (approved):
Looks good to me, I'm approving it!

PR #1176, opened 2025-01-02
solution

yuliiazubenko:
[DEMO LINK](https://yuliiazubenko.github.io/react_people-table-advanced/)


save-v (approved):
Good job! However, pay attention that your 'All' button resets all filters and sorting parameters, but it should only reset the century filter

PR #1178, opened 2025-01-03
Develop

Andrii-Yanchuk:
[DEMO LINK](https://Andrii-Yanchuk.github.io/react_people-table-advanced/)


src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    setSearchParams(search);
+  }
+
+  return (
+    <nav className="panel">
+      <p className="panel-heading">Filters</p>
+
+      <p className="panel-tabs" data-cy="SexFilter">
+        {Object.values(PersonSex).map(sexFilter => {
```
AnderZerfall, line 24: That's not a big deal, but you could create an enum Sex { All = '', Female = 'f', Male = 'm' } and use Object.keys() to create the same tabs from the example.  A key would play a role of tab's title (All. Female. Male) and key's value would be a search parameter ('', 'f', 'm').

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,110 @@
...
+            <i className="fas fa-search" aria-hidden="true" />
+          </span>
+        </p>
+      </div>
+
+      <div className="panel-block">
+        <div className="level is-flex-grow-1 is-mobile" data-cy="CenturyFilter">
+          <div className="level-left">
+            {CENTURIES.map(century => {
+              const currentCenturies = centuries.includes(century)
```
AnderZerfall, line 67: There is too much code here. I believe it would make sense to move some part of it to a separate function. To be more specific, the part where you filter your centuries array.

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,91 @@
...
+      <h1 className="title">People Page</h1>
+      <div className="block">
+        <div className="columns is-desktop is-flex-direction-row-reverse">
+          <div className="column is-7-tablet is-narrow-desktop">
+            {!!peoples.length && !isLoading && <PeopleFilters />}
+          </div>
+
+          <div className="column">
+            <div className="box table-container">
+              {isLoaderVisible && <Loader />}
```
AnderZerfall, line 66: So, I noticed one small issue here. The message 'There are no people matching the current search criteria' appears with the loader.  The thing is, that you haven't received any data from the server yet, 'cause the loading is still ongoing. 
So, whenever you show the loader, hide everything else. Display this error message only if the timeout was reached and the server hasn't returned anything.

image

AnderZerfall (approved):
Good job! Though, I've listed some moments that should be noted

PR #1188, opened 2025-01-05
solution

VitaliiShc:
[DEMO LINK](https://VitaliiShc.github.io/react_people-table-advanced/)


pogorielova (approved):
Great work! But you should also implement sorting buttons to change: those arrows should change according to the sorting order

PR #1253, opened 2025-02-22
Task solution

Dominik-Dudek-code:
- [DEMO LINK](https://Dominik-Dudek-code.github.io/react_people-table-advanced/)


Dominik-Dudek-code:
I would like to understand this module even better, and I prefer learning through videos. Could you recommend any reliable sources where I can gain more knowledge? Since there were no explanatory videos from Bartek included in this module, it's not entirely clear whether I did everything in the best way.

danon321 (approved):
Dobra robota. Jak na moje wszystko śmiga poprawnie

PR #1303, opened 2025-04-05
Develop

Artem-Brui:
[DEMO LINK](https://Artem-Brui.github.io/react_people-table-advanced/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. Remove this block 
image

2. This block looks different as on the working example, need to fix it. You also need to show this block only after people have loaded it.
image

3. I can't type any symbol in the search input 
image

src/components/Error.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Loader } from './Loader';
+
+export const Error = () => {
+  const {
+    context: { isLoading, error },
+  } = useContext(Context);
+
+  const isVisible = isLoading || error;
+
+  return isVisible ? (
```
etojeDenys, line 12:
```suggestion
  if (isLoading || error) {
   return null;
  }

  return (
```
looks better

src/components/Navbar.tsx
```diff
@@ -1,3 +1,14 @@
+import { NavLinkItem } from './NavLinkItem';
+
+export enum NavLinks {
+  Home,
+  People,
+}
+
+const Links = Object.keys(NavLinks).filter(key =>
```
etojeDenys, line 8:
```suggestion
const links = Object.keys(NavLinks).filter(key =>
```

src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const PeopleFilters = () => {
+  const [searchParams] = useSearchParams();
+  const [query, setQuery] = useState(searchParams.get('query') || '');
+  const {
+    context: { isLoading },
+  } = useContext(Context);
+
+  const isVisible = !isLoading;
+
+  return isVisible ? (
```
etojeDenys, line 18: same here

src/components/PeopleFilters/components/CenturyFilter.tsx
```diff
@@ -0,0 +1,59 @@
+import classNames from 'classnames';
+import { useSearchParams } from 'react-router-dom';
+import React, { useCallback } from 'react';
+import { getSearchWith } from '../../../utils/searchHelper';
+import { updateSearchParams } from '../../PeopleTable/utils/updateSearchParams';
+
+const CenturyFilters = [16, 17, 18, 19, 20];
```
etojeDenys, line 7:
```suggestion
const CenturyFilters = [16, 17, 18, 19, 20];
```
in camelCase or in UPPER_CASE if it's constant

etojeDenys (changes requested):
centuries looks broken, if I select 16 it shows the same amount of people as all selected
image

volodymyr-soltys97 (approved):
Good job 👍

PR #1306, opened 2025-04-07
add task solution

Hy-tapa-kot:
[DEMO LINK](https://Hy-tapa-kot.github.io/react_people-table-advanced/)


etojeDenys (changes requested):
try to pass all tests and then re-request the review

etojeDenys (changes requested):
doesn't work now at all
image

PR #1313, opened 2025-04-09
add solution

daniil-trusov:
[DEMO LINK](https://daniil.trusov.github.io/react_people-table-advanced/)


src/Root.tsx
```diff
@@ -0,0 +1,33 @@
...
+        <Route path="*" element={<NotFoundPage />} />
+
+        <Route path="people">
+          <Route
+            path=":activePerson?"
+            element={
+              <PeopleContextProvider>
+                <PeoplePage />
+              </PeopleContextProvider>
+            }
```
AnOvramenko, line 27: I asked about how to wrap component in Route by Context, as I understand right it is better to create additional component like PeoplePageWithContext and put it in element attribute, but I think it is also a nice solution

AnOvramenko (approved):
Good job! 👍

PR #1322, opened 2025-04-10
add solution

R-Ohirok:
[DEMO LINK](https://R-Ohirok.github.io/react_people-table-advanced/)


src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,41 @@
...
+          <NavLink
+            to="/people"
+            className={({ isActive }) => {
+              return cn('navbar-item', {
+                'has-background-grey-lighter': isActive,
+              });
+            }}
+          >
+            People
+          </NavLink>
```
vladrlex, line 36: Looks like those two components use the same logic in className, what do you think about cut it to separate callback for reusing. Because at the moment you have only two navigation links but for scaleability it will be better to move it to reuseable function.

src/functions/FilterPeople.ts
```diff
@@ -0,0 +1,70 @@
+import { useMemo } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import { Person } from '../types';
+import { paramsForSearch } from '../utils/paramsForSearch';
+import { filterByNames } from './FilterByNames';
+import { filterBySex } from './FilterBySex';
+import { filterByCenturies } from './FilterByCenturies';
+import { SortParams } from '../types/sortParams';
+
+export const FilterPeople = (people: Person[]): Person[] => {
```
vladrlex, line 10: I think it should be `useFilterPeople` - custom hooks names must always start with `use` prefix.

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,85 @@
...
+            <div className="column">
+              <div className="box table-container">
+                <Loader />
+              </div>
+            </div>
+          </div>
+        </div>
+      </>
+    );
+  }
```
vladrlex, line 53: It can be inside of return, here you repeat yourself because only `Loader` component added in condition

src/functions/FilterPeople.ts
```diff
@@ -0,0 +1,70 @@
...
+      filteredPeople = filterBySex(filteredPeople, sexFilter);
+    }
+
+    if (normalisedQuery) {
+      filteredPeople = filterByNames(filteredPeople, normalisedQuery);
+    }
+
+    if (!!centuryFilters.length) {
+      filteredPeople = filterByCenturies(filteredPeople, centuryFilters);
+    }
```
vladrlex, line 37: If a user visits the website using a link with search parameters, the filtering function will be called three times.
Currently, the people array is quite small, but imagine it contains 1000 users — in that case, the callback inside each filter will be executed too many times.
My suggestion is to combine the filtering logic into a single function that checks all three conditions at once.

vladrlex (approved):
Well done, great job! Just a few small remarks.
Also, inside the functions folder, please rename the files so that their names start with a lowercase letter.

PR #1329, opened 2025-04-12
task solution

VladRoz95:
[DEMO LINK](https://VladRoz95.github.io/react_people-table-advanced/)


nataliia-v (approved):
One issue, is that when there are no people matching the current search criteria, we shouldn't render the table header.

image

PR #1342, opened 2025-04-17
☼ution

ENigmusSS:
[DEMO LINK](https://enigmusss.github.io/react_people-table-advanced/)


ENigmusSS:
@mateacademy-ai-mentor 
The has() method of the [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) interface returns a boolean value that indicates whether the specified parameter is in the search parameters.

A parameter name and optional value are used to match parameters. If only a parameter name is specified, then the method will return true if any parameters in the query string match the name, and false otherwise. If both a parameter name and value are specified, then the method will return true if a parameter matches both the name and value. Source: MDN Web Docs

Switch case always take the first true case, so if isFilteredOut is true, no matter what value isEmpty has

PR #1376, opened 2025-05-24
commit 1.0

KvashaIhor:
[DEMO LINK](https://kvashaihor.github.io/react_people-table-advanced/)

### DIFF
--- a/src/components/Navbar.tsx
+++ b/src/components/Navbar.tsx
@@ -1,4 +1,8 @@
+import { Link, useLocation } from 'react-router-dom';
+
 export const Navbar = () => {
+  const location = useLocation();
+
   return (
     <nav
       data-cy="nav"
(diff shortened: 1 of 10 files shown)


src/components/Navbar.tsx
```diff
@@ -8,17 +12,20 @@ export const Navbar = () => {
     >
       <div className="container">
         <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
+          <Link
+            className={`navbar-item ${location.pathname === '/' || location.pathname === '' ? 'navbar-item has-background-grey-lighter' : ''}`}
```
volodymyr-soltys97, line 16: use classnames library for add classes with condition

src/components/PeopleFilters.tsx
```diff
@@ -1,16 +1,46 @@
...
+  clearCenturies,
+}: FilterProps) => {
   return (
     <nav className="panel">
       <p className="panel-heading">Filters</p>
 
       <p className="panel-tabs" data-cy="SexFilter">
-        <a className="is-active" href="#/people">
+        <a
+          className={!sex ? 'is-active' : ''}
```
volodymyr-soltys97, line 26: the same as above, fix it everywhere

src/components/PeopleFilters.tsx
```diff
@@ -35,40 +67,70 @@ export const PeopleFilters = () => {
           <div className="level-left">
             <a
               data-cy="century"
-              className="button mr-1"
-              href="#/people?centuries=16"
+              className={
+                centuries.includes(16) ? 'button mr-1 is-info' : 'button mr-1'
+              }
+              href="#"
```
volodymyr-soltys97, line 73: create array for centuries and render these filters with map() method

volodymyr-soltys97 (changes requested):
Good job
To improve 
1. Need to show this block only after people loaded
image

2. Don't need to reset selected filters after clicking on the person
image
image
image

PR #1418, opened 2025-06-26
RAW solution. Will fix with next commit

jasialazyrka:
[DEMO LINK](https://jasialazyrka.github.io/react_people-table-advanced/)


YarRudenko (changes requested):
your code looks great, but you need to implement sorting, query and filtering by sex

PR #1455, opened 2025-07-24
Develop

14lucas-mendes:
[DEMO LINK](https://14lucas-mendes.github.io/react_people-table-advanced/)


DanielCamposDevX (changes requested):
Centuries filter should be a grouped filter not just one at a time.

PR #1519, opened 2025-09-10
solution

DidZhara:
[DEMO LINK](https://DidZhara.github.io/react_people-table-advanced/)


src/components/Navbar.tsx
```diff
@@ -8,17 +20,19 @@ export const Navbar = () => {
...
+          </Link>
 
-          <a
-            aria-current="page"
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
+          <Link
+            className={`navbar-item ${isPeopleActive ? 'has-background-grey-lighter' : ''}`}
+            to={peopleLink}
           >
```
derpedcatto, line 33: `NavLink` components are better suited for navigation like this without extra fluff. Here is an example from my code:
```tsx
export const NavBar = () => {
  return (
    <nav
      className="navbar is-fixed-top has-shadow"
      data-cy="nav"
      role="navigation"
      aria-label="main navigation"
    >
      <div className="container">
        <ul className="navbar-brand">
          <li>
            <NavLink
              to="/"
              className={({ isActive }) =>
                `navbar-item ${isActive ? 'has-background-grey-lighter' : ''}` // why did I write it like this bruh
              }
              end
            >
              Home
            </NavLink>
          </li>

          <li>
            <NavLink
              to="/people"
              className={({ isActive }) =>
                `navbar-item ${isActive ? 'has-background-grey-lighter' : ''}` // why did I write it like this bruh
              }
            >
              People
            </NavLink>
          </li>
        </ul>
      </div>
    </nav>
  );
};
```

src/components/PeoplePage.tsx
```diff
@@ -1,33 +1,158 @@
...
+    getPeople()
+      .then(data => {
+        setPeople(data);
+        setLoading(false);
+      })
+      .catch(() => {
+        setError(true);
+        setLoading(false);
+      });
+  }, []);
```
derpedcatto, line 29: All is correct, but you could condense this a little bit by moving `setLoading(false);` to `finally` block after `catch`. 

Also here is an example of how you could write this using `try/catch`:
```tsx
  const handleLoadPeople = async () => {
    try {
      const apiPeople = await getPeople();

      setPeople(apiPeople);
    } catch (error) {
      setIsError(true);
    }

    setIsLoading(false);
  };

  useEffect(() => {
    handleLoadPeople();
  }, []);
```

src/components/PeoplePage.tsx
```diff
@@ -1,33 +1,158 @@
...
+      const lowerQuery = query.toLowerCase();
+
+      filteredPeople = filteredPeople.filter(
+        person =>
+          person.name.toLowerCase().includes(lowerQuery) ||
+          (person.motherName &&
+            person.motherName.toLowerCase().includes(lowerQuery)) ||
+          (person.fatherName &&
+            person.fatherName.toLowerCase().includes(lowerQuery)),
+      );
```
derpedcatto, line 46: You could condense this a little bit like this:
```ts
    .filter(person => {
      return (
        person.name.toLowerCase().includes(query) ||
        person.motherName?.toLowerCase().includes(query) ||
        person.fatherName?.toLowerCase().includes(query)
      );
    })
```

src/components/PeoplePage.tsx
```diff
@@ -1,33 +1,158 @@
...
+        if (aValue < bValue) {
+          return order === 'desc' ? 1 : -1;
+        }
+
+        if (aValue > bValue) {
+          return order === 'desc' ? -1 : 1;
+        }
+
+        return 0;
+      });
```
derpedcatto: Here is an alternative version of how this could be written:
```ts
  if (sort) {
    return filteredPeople.sort((a, b) => {
      let comparison = 0;

      const fieldA = a[sort as keyof Person];
      const fieldB = b[sort as keyof Person];

      if (typeof fieldA === 'string' && typeof fieldB === 'string') {
        comparison = fieldA.localeCompare(fieldB);
      }

      if (typeof fieldA === 'number' && typeof fieldB === 'number') {
        comparison = fieldA - fieldB;
      }

      return order === 'desc' ? -comparison : comparison;
    });
  }
```

src/components/PeoplePage.tsx
```diff
@@ -1,33 +1,158 @@
...
+        <p data-cy="peopleLoadingError" className="has-text-danger">
+          Something went wrong
+        </p>
+      )}
+
+      {!loading && !error && people.length === 0 && (
+        <p data-cy="noPeopleMessage">There are no people on the server</p>
+      )}
+
+      {!loading && !error && people.length > 0 && (
```
derpedcatto, line 146: To not deal with repeating checks like `!loading` you could write a separate function like this:
```tsx
  const renderContent = () => {
    if (isError) {
      return (
        <p data-cy="peopleLoadingError" className="has-text-danger">
          Something went wrong
        </p>
      );
    }

    if (people.length === 0) {
      return <p data-cy="noPeopleMessage">There are no people on the server</p>;
    }

    if (preparedPeople.length === 0) {
      return <p>There are no people matching the current search criteria</p>;
    }

    return (
        <PeopleTable people={preparedPeople} selectedPersonSlug={personSlug} />
      </div>
    );
  };

  return (
    <>

      <div className="block">
          {isLoading ? (
            <Loader />
          ) : (
            <>
                <PeopleFilters />
              </div>

              <div className="column">{renderContent()}</div>
            </>
          )}
        </div>
      </div>
    </>
  );
```

src/components/PeopleTable.tsx
```diff
@@ -1,645 +1,142 @@
...
+                </SearchLink>
+              </th>
+              <th>
+                <SearchLink params={getSortParams('died')}>
+                  Died{getSortIcon('died')}
+                </SearchLink>
+              </th>
+              <th>Mother</th>
+              <th>Father</th>
+            </tr>
```
derpedcatto, line 83: Usage of arrow symbols is incorrect. You should use icons instead. Take another look at working example and see how it is different.

Here are their classnames:
- Default - `fa-sort`
- Down - `fa-sort-down`
- Up - `fa-sort-up`

Here is an example of how it should look like:
```tsx
    <th>
      <span className="is-flex is-flex-wrap-nowrap">
        {title}
        <SearchLink params={params}>
          <span className="icon">
            <i className={classNames('fas', iconClassName)} /> // iconClassName - default, down or up className
          </span>
        </SearchLink>
      </span>
    </th>
```

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  const linkClass = person.sex === 'f' ? 'has-text-danger' : '';
+  const linkTo = `/people/${person.slug}?${searchParams.toString()}`;
+
+  return (
+    <Link to={linkTo} className={linkClass}>
+      {name}
+    </Link>
+  );
```
derpedcatto, line 20: Here is an alternative way to write this link:
```tsx
    <Link
      to={{
        pathname: `./${person.slug}`,
        search: searchParams.toString(),
      }}
      className={classNames({
        'has-text-danger': person.sex === 'f',
        'has-text-info': person.sex === 'm',
      })}
    >
      {person.name}
    </Link>
```

derpedcatto (approved):
Good job! The main thing that should probably be fixed is using icons instead of arrow symbols in `PeopleTable`, but if AI passes your version then whatever. Other than that everything else looks good I think.

I left extra comments about how you could refactor some places in code, you can check them out if you want ✌️

PR #1522, opened 2025-09-11
Develop

Oleksandr-Oliferuk:
[DEMO LINK](https://Oleksandr-Oliferuk.github.io/react_people-table-advanced/)


src/components/PeopleFilters.tsx
```diff
@@ -35,10 +43,16 @@ export const PeopleFilters = () => {
       <div className="panel-block">
         <p className="control has-icons-left">
           <input
+            value={query ?? ''}
             data-cy="NameFilter"
             type="search"
             className="input"
             placeholder="Search"
+            onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+              const inputQuery = event.target.value;
```
DmitriyALEX: You can get event.target.value in html and in function parameters use only string instead React.ChangeEvent<HTMLInputElement>

src/components/Table/Table.tsx
```diff
@@ -19,42 +43,96 @@ export const Table: React.FC<Props> = ({ people, isLoading }) => {
...
-                  <a href="#/people?sort=name">
+                  <SearchLink
+                    params={getPrepareSortParams(sort, order, 'name')}
+                  >
+                    {' '}
                     <span className="icon">
-                      <i className="fas fa-sort" />
+                      <i
+                        className={cn(
+                          'fas',
```
DmitriyALEX, line 53: get this logic outside html to function which will get parametr like 'name' or 'sex' and call this function in cn like 

function getStyles = (sortBy: string) => {
//logic
}

cn(getStyles('name')

DmitriyALEX (approved):
good job :)

PR #1526, opened 2025-09-11
solution

iastivka:
[DEMO LINK](https://iastivka.github.io/react_people-table-advanced/)


lvlashnin (approved):
everything works as it should The only thing I don't like is that parameters from the URL can be obtained from the parent component trough useSerchParams and no need pass them through props

PR #1580, opened 2025-11-03
add tack complited

Mishadoloh:
[DEMO LINK](https://Mishadoloh.github.io/react_people-table-advanced/)


src/components/PeopleFilters/PeopleFilters.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { Sex } from '../../types/Sex';
+
+const CENTURIES_ARRAY = [16, 17, 18, 19, 20];
+
+export const PeopleFilters = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const query = searchParams.get('query') || '';
+  const centuries = searchParams.getAll('centuries') || [];
+
+  function handleQueryChange(event: React.ChangeEvent<HTMLInputElement>) {
```
volodymyr-soltys97: It is important to follow the same way of creating functions, fix it everywhere
```suggestion
  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```

volodymyr-soltys97 (changes requested):
Hi, you need to fix the demo link when we don't have people
Demo
image


Working example 
image

volodymyr-soltys97 (changes requested):
Hi, this comment still not fixed from the previous review, you need to hide the table header when you don't have people. Check it on the working example
image

volodymyr-soltys97 (changes requested):
Hi, the demo link still not fixed from the previous review, you need to make a deploy again after your changes and chect the demo link after 
image

vitalii-bondarenko-ua (approved):
Looks like fixed.
Good job.

PR #1596, opened 2025-11-12
add task solution

Julia-Hertseva:
[DEMO LINK](https://Julia-Hertseva.github.io/react_people-table-advanced/)


src/components/PeopleFilters.tsx
```diff
@@ -33,61 +79,37 @@ export const PeopleFilters = () => {
...
-              href="#/people?centuries=20"
-            >
-              20
-            </a>
+            {CENTURIES.map(century => (
+              <SearchLink
+                key={century}
+                data-cy="century"
+                className={getCenturyButtonClass(century)}
+                params={{ centuries: getCenturyParams(century) }}
```
pshy4ok, line 87: I think you should select centuries and sex with onClick

pshy4ok (changes requested):
Good job! But let's improve your solution. In your project, filtering by gender and century does not work, and there is also a problem with displaying the loader when going to the PeoplePage.

PR #1607, opened 2025-11-13
Add task solution

Volodymyr-Bauzi:
[DEMO LINK](https://Volodymyr-Bauzi.github.io/react_people-table-advanced/)


dashasssss (changes requested):
Your solution is missing the implementation of sorting by centuries, and the sorting logic with the arrows in the table is also not implemented.

PR #1683, opened 2026-02-11
Develop

illia-001:
[DEMO LINK](https://illia-001.github.io/react_people-table-advanced/)


src/App.tsx
```diff
@@ -10,9 +13,15 @@ export const App = () => {
 
       <div className="section">
         <div className="container">
-          <h1 className="title">Home Page</h1>
-          <h1 className="title">Page not found</h1>
-          <PeoplePage />
+          <Routes>
+            <Route path="/" element={<HomePage />} />
+            <Route path="/home" element={<Navigate to="/" replace />} />
+            <Route path="*" element={<PageNotFound />} />
```
dmytro-halieba: I'd move this path after the people route for better visual grouping and clarity.

src/utils/getFilteredPeople.ts
```diff
@@ -0,0 +1,57 @@
...
+    );
+  }
+
+  if (filterBySex) {
+    newPeople = newPeople.filter(person => person.sex === filterBySex);
+  }
+
+  const direction = searchOrder === SortOrders.Descedent ? -1 : 1;
+
+  newPeople.sort((a: Person, b: Person) => {
```
dmytro-halieba: Also consider wrapping the sort in a conditional check `(if (searchOption))` since we currently sort unnecessarily even when not needed.

src/utils/getFilteredPeople.ts
```diff
@@ -0,0 +1,57 @@
+import { Person } from '../types';
+import { SortOptions } from '../types/SortOptions';
+import { SortOrders } from '../types/SortOrders';
+
+export const getFilteredPeople = (
```
dmytro-halieba, line 7: The function name getFilteredPeople is misleading - it also sorts the data. Consider renaming it

src/components/PeoplePage.tsx
```diff
@@ -1,29 +1,54 @@
...
             <div className="box table-container">
-              <Loader />
-
-              <p data-cy="peopleLoadingError">Something went wrong</p>
-
-              <p data-cy="noPeopleMessage">There are no people on the server</p>
-
-              <p>There are no people matching the current search criteria</p>
+              {errorMessage && (
+                <p data-cy={handleErrorMessage(errorMessage)}>{errorMessage}</p>
```
dmytro-halieba, line 46: That works, but I’m not a big fan of using errorMessage for all these cases.

Conceptually, only “Something went wrong” is an actual error.
Messages like “There are no people on the server” or “There are no people matching the current search criteria” represent valid empty states, not errors.

src/components/PeopleLink.tsx
```diff
@@ -0,0 +1,61 @@
...
+        >
+          {name}
+        </Link>
+      </td>
+      <td>{sex}</td>
+      <td>{born}</td>
+      <td>{died}</td>
+      <td>
+        {(mother ? (
+          <ParentLink parent={mother} search={search} />
```
dmytro-halieba: This is not an issue, but I think this logic could be nicely (for line 35, 47 and 54) extracted into a reusable component (e.g. PersonLink).

Right now, the same rendering pattern is used for a person, mother, and father (link if the entity exists, fallback to name or - otherwise, plus styling logic).

Extracting this into a small component would help reduce duplication, improve readability, and make future changes easier (e.g. if the link styling or behavior changes, it will be updated in one place).

dmytro-halieba (approved):
I tried to be quite picky in my review to point out potential improvements for the future. Overall, your code is very well done the structure is clear, and you’ve done a great job splitting logic into reusable components and types.

One small thing to note: as I can see, you didn’t use the `SearchLink` component that was provided at the beginning to simplify working with search params. I’m not sure if that was intentional, but if not, I’d recommend taking a look at its functionality and considering replacing Link with SearchLink where appropriate.

Keep up the great work - this is a solid and well-structured implementation

PR #1769, opened 2026-04-29
Implement the ability to filter and sort people in the table

Dianaeva:
[DEMO LINK](https://dianaeva.github.io/react_people-table-advanced/)


src/components/SortLink/SortLink.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const sortParam = searchParams.get(URLSearchParam.Sort);
+  const orderParam = searchParams.get(URLSearchParam.Order);
+
+  const getSortSearchWith = () => {
+    if (sortParam !== sortField) {
+      return getSearchWith(searchParams, {
+        [URLSearchParam.Sort]: sortField,
+        [URLSearchParam.Order]: null,
+      });
+    } else if (sortParam === sortField && !orderParam) {
```
polosanya:
```suggestion
    if (sortParam !== sortField) {
      return getSearchWith(searchParams, {
        [URLSearchParam.Sort]: sortField,
        [URLSearchParam.Order]: null,
      });
    }
    
     if (sortParam === sortField && !orderParam) {
```
There is no need for `else` after `return`. Also, would be better to use `switch..case` here and just change the order after it if needed

polosanya (changes requested):
Almost done! Pay attention to the comments from previous task too

polosanya (approved):

PR #1797, opened 2026-06-01
task solution

DmitriyS2k:
[DEMO LINK](https://DmitriyS2k.github.io/react_people-table-advanced/)


src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+          default:
+            return 0;
+        }
+      });
+
+      if (order === 'desc') {
+        filteredPeople.reverse();
+      }
+    }
```
dothearts, line 171: It can be pure function, which you move outside component, import and use in this place.

dothearts (approved):

PR #1803, opened 2026-06-03
feat: implement the ability to filter and sort people in the table

linska:
[demo](https://linska.github.io/react_people-table-advanced/)


Evgeniy45 (changes requested):
I noticed a couple of issues with how the filters and rendering work:
Century filter: The 'All' button should be active by default and become inactive when a specific century is selected. Currently, it works the exact opposite way — selecting a century makes the 'All' button active.
Loading state: The table and filters should render together after the loading phase finishes. Right now, the filters appear first, followed by the table.

Could you please fix these? :)

PR #1811, opened 2026-06-05
Sol

sergeisvidlo4:
[DEMO](https://sergeisvidlo4.github.io/react_people-table-advanced/)


dothearts (changes requested):
In general, all fine. 

But male/female filter doesn't work currently. Please fix it! 

Image

Many thanks!

dothearts (approved):

